PR #2, opened 2021-05-26
Add missing tests

jbajic:
Add tests for:
 - `utilities.py`
 - memgraph mapping for Node, Relationship and Path

### DIFF
--- a/gqlalchemy/models.py
+++ b/gqlalchemy/models.py
@@ -111,8 +111,8 @@ class Relationship(UniqueGraphObject):
         self,
         rel_id: Any,
         rel_type: str,
-        start_node: Node,
-        end_node: Node,
+        start_node: int,
+        end_node: int,
         properties: Dict[str, Any] = None,
     ):
         super().__init__(rel_id, properties)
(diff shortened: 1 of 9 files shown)


gqlalchemy/models.py
```diff
@@ -111,8 +111,8 @@ def __init__(
         self,
         rel_id: Any,
         rel_type: str,
-        start_node: Node,
-        end_node: Node,
+        start_node: int,
```
jmatak, line 114: Why should nodes be expressed as integers?

jbajic: Because they are integers, when returning edges you get start_node and end_node as memgraph's internal ids

jbajic: It would make sense to implement lazy fetching when the user refers to the start/end node we would fetch it from memgraph (pro: you actually get what you want, con: getters could break). What do yous say?

jmatak: Do you think that it would lead toward a high number of database interactions? I imagine multiple users instantly iterating over edges and fetching nodes (because edge information itself is useless without starting and ending nodes).

jbajic: Yes it definitely would be an overhead, but there are two ways to solve this (as far as I know):
- We can load start/end node lazily when there are being fetched
- When executing a query that fetches edge, change it to fetch nodes as well and save them in edge data
I would leave it as is for now. since the user gets the start and end node in the same row, and can easily say which are start and end nodes.

jmatak (approved):
:+1:

Josipmrden (approved):
Great :)

PR #3, opened 2021-05-26
Add networkx transformation

jbajic:
Integrate networkx graph to cypher queries transformation.
### How to use?

Currently there are two new API calls to `gqlalchemy.tranformations`:
 - `nx_to_cypher`
 - `nx_graph_to_memgraph_parallel`

####  `nx_to_cypher`
 Accepts a networkx graph and translates it into two (or three) groups of cypher queries. The first group is a creation of nodes, second is optional (if the caller passes `create_index` as `True`) you will get queries for creating indexes on `id` property, and this is a creation of edges.

####  `nx_graph_to_memgraph_parallel`
Accepts a networkx graph and connection parameters for creating multiple connections to memgraph which will be used to insert the whole networkx graph in memgraph in parallel. It also supports `create_index` flag, if set the function will create all indexes an `id` property.

Using this module assumes some stuff:
 - If you create nodes with the property `labels` it will be taken as a label and will not be a part of the properties.
 - If you create edges with the property `type` will be taken as a type for node and will not be a part of properties
 - The `id` that networx has for nodes will be set ad id in the node properties.

### DIFF
--- a/gqlalchemy/transformations.py
+++ b/gqlalchemy/transformations.py
@@ -1,0 +1,63 @@
+# Copyright (c) 2016-2021 Memgraph Ltd. [https://memgraph.com]
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, Dict, Iterator
+
+import networkx as nx
+
+from gqlalchemy.utilities import to_cypher_labels, to_cypher_properties
+
+__all__ = (
+    "nx_to_cypher",
+    "nx_edges_to_cypher",
+    "nx_nodes_to_cypher",
+)
+
+
+def nx_nodes_to_cypher(graph: nx.Graph) -> Iterator[str]:
+    """Generates a Cypher queries for creating nodes"""
+    for nx_id, data in graph.nodes(data=True):
+        yield _create_node(nx_id, data)
+
+
+def nx_edges_to_cypher(graph: nx.Graph) -> Iterator[str]:
+    """Generates a Cypher queries for creating edges"""
+    for n1, n2, data in graph.edges(data=True):
+        yield _create_edge(n1, n2, data)
+
+
+def nx_to_cypher(graph: nx.Graph) -> Iterator[str]:
+    """Generates a Cypher queries for creating graph"""
+
+    yield from nx_nodes_to_cypher(graph)
+    yield from nx_edges_to_cypher(graph)
+
+
+def _create_node(nx_id: int, properties: Dict[str, Any]) -> str:
+    """Returns Cypher query for node creation"""
+    if "id" not in properties:
+        properties["id"] = nx_id
+    labels_str = to_cypher_labels(properties.pop("labels", ""))
+    properties_str = to_cypher_properties(properties)
+
+    return f"CREATE ({labels_str} {properties_str});"
+
+
+def _create_edge(from_id: int, to_id: int, properties: Dict[str, Any]) -> str:
+    """Returns Cypher query for edge creation."""
+    edge_type = to_cypher_labels(properties.get("type", "TO"))
+    properties.pop("type", None)
+    properties_str = to_cypher_properties(properties)
+
+    return f"MATCH (n {{id: {from_id}}}), (m {{id: {to_id}}}) CREATE (n)-[{edge_type} {properties_str}]->(m);"
--- a/gqlalchemy/transformations.py
+++ b/gqlalchemy/transformations.py
@@ -121,8 +121,6 @@ def _insert_queries(queries: List[str], host: str, port: int, username: str, pas
         try:
             query = queries.pop()
             memgraph.execute_query(query)
-        except IndexError:
-            break
         except mgclient.DatabaseError as e:
             queries.append(query)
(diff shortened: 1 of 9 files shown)


gqlalchemy/transformations.py
```diff
@@ -0,0 +1,63 @@
...
+    return f"CREATE ({labels_str} {properties_str});"
+
+
+def _create_edge(from_id: int, to_id: int, properties: Dict[str, Any]) -> str:
+    """Returns Cypher query for edge creation."""
+    edge_type = to_cypher_labels(properties.get("type", "TO"))
+    properties.pop("type", None)
+    properties_str = to_cypher_properties(properties)
+
+    return f"MATCH (n {{id: {from_id}}}), (m {{id: {to_id}}}) CREATE (n)-[{edge_type} {properties_str}]->(m);"
```
gitbuda, line 63: This will be slow. I would force users to define a Label and then `CREATE INDEX ON :Label(id);`?

gqlalchemy/transformations.py
```diff
@@ -0,0 +1,152 @@
...
+def nx_to_cypher(graph: nx.Graph) -> Iterator[str]:
+    """Generates a Cypher queries for creating graph"""
+
+    yield from _nx_nodes_to_cypher(graph)
+    yield from _nx_edges_to_cypher(graph)
+
+
+def nx_graph_to_memgraph_parallel(
+    graph: nx.Graph,
+    host: str = "127.0.01",
```
gitbuda, line 43: missing `.` :smile:

gqlalchemy/transformations.py
```diff
@@ -0,0 +1,152 @@
...
+):
+    memgraph = Memgraph(host, port, username, password, encrypted)
+    indexes = memgraph.get_indexes()
+    if len(indexes) == 0:
+        logging.getLogger(__file__).warning(
+            f"Be careful you do not have any indexes set up, the queries will take longer than expected!"
+        )
+
+
+def _insert_queries(queries: List[str], host, port, username, password, encrypted) -> None:
```
MasterMedo, line 98: No type hints for arguments?

gqlalchemy/transformations.py
```diff
@@ -0,0 +1,152 @@
...
+    memgraph = Memgraph(host, port, username, password, encrypted)
+    while len(queries) > 0:
+        try:
+            query = queries.pop()
+            memgraph.execute_query(query)
+        except IndexError:
+            break
+        except mgclient.DatabaseError as e:
+            queries.append(query)
+            logging.getLogger(__file__).warning(f"Ignoring database error: {str(e)}")
```
MasterMedo, line 109: call to `str` in an f-string? Does it make a difference?

MasterMedo (changes requested):

gqlalchemy/transformations.py
jbajic: Nope it does not will remove

gqlalchemy/transformations.py
```diff
@@ -0,0 +1,152 @@
...
+    port: int = 7687,
+    username: str = "",
+    password: str = "",
+    encrypted: bool = False,
+):
+    memgraph = Memgraph(host, port, username, password, encrypted)
+    indexes = memgraph.get_indexes()
+    if len(indexes) == 0:
+        logging.getLogger(__file__).warning(
+            f"Be careful you do not have any indexes set up, the queries will take longer than expected!"
```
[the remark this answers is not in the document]
jbajic: Will remove

gqlalchemy/transformations.py
```diff
@@ -0,0 +1,152 @@
...
+        username,
+        password,
+        encrypted,
+    )
+    for queries_gen in [_nx_nodes_to_cypher(graph), _nx_edges_to_cypher(graph)]:
+        queries = list(queries_gen)
+        chunk_size = len(queries) // num_of_processes
+        processes = []
+        for i in range(num_of_processes):
+            process_queries = queries[i * chunk_size : chunk_size * (i + 1)]
```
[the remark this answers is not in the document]
jbajic: Black add these automatically, both the whitespace between `:` and `+`. I would not change it since its being formatted automatically

gqlalchemy/transformations.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def nx_to_cypher(graph: nx.Graph, create_index=False) -> Iterator[str]:
+    """Generates a Cypher queries for creating graph"""
+
+    if create_index:
+        yield from _nx_nodes_to_cypher_with_index(graph)
+    else:
+        yield from _nx_nodes_to_cypher(graph)
+    yield from _nx_edges_to_cypher(graph)
```
Josipmrden, line 42: Why would you want to yield either index OR node creation?

gqlalchemy/transformations.py
```diff
@@ -0,0 +1,192 @@
...
+    while len(queries) > 0:
+        try:
+            query = queries.pop()
+            memgraph.execute_query(query)
+        except IndexError:
+            break
+        except mgclient.DatabaseError as e:
+            queries.append(query)
+            logging.getLogger(__file__).warning(f"Ignoring database error: {e}")
+            continue
```
Josipmrden, line 129: Why not log the IndexError as well. When will it happen?

jbajic: Well technically it should never happen, that why I am going to remove the catch

Josipmrden (approved):

gqlalchemy/transformations.py
```diff
@@ -0,0 +1,190 @@
...
+        query_groups.append(_nx_nodes_to_cypher_with_index(graph))
+    else:
+        _check_for_index_hint(
+            host,
+            port,
+            username,
+            password,
+            encrypted,
+        )
+        query_groups.append(_nx_nodes_to_cypher(graph))
```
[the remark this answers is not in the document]
jbajic: It is if you set `create_index` flag to `True`

jmatak (approved):

MasterMedo (approved):

PR #7, opened 2021-06-04
Fix badge urls

jbajic:


jmatak (approved):
It works for images in row, nice!

PR #10, opened 2021-06-28
Update workflow

jbajic:
Update workflow to test library on all supported python versions

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -10,7 +10,7 @@ jobs:
   build_and_test_ubuntu:
     strategy:
       matrix:
-        python-version: ['3.8', '3.9']
+        python-version: [ '3.8', '3.9' ]
         os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -13,11 +13,9 @@ jobs:
         python-version: ['3.8', '3.9']
         os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
-
     steps:
     - name: Checkout Repository
       uses: actions/checkout@v2
-
     - name: Set up Python ${{ matrix.py-version }}
       uses: actions/setup-python@v2
       with:
@@ -26,7 +24,6 @@ jobs:
       uses: abatilo/actions-poetry@v2.0.0
       with:
         poetry-version: ${{ env.POETRY_VERSION }}
-
     - name: Run memgraph instance
       run: |
         docker run -d -p 7687:7687 memgraph/memgraph --telemetry-enabled=False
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -1,33 +1,66 @@
...
 on: [push]
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
-
+  build_and_test_ubuntu:
+    strategy:
+      matrix:
+        python-version: ['3.8', '3.9']
```
jmatak, line 13: This array is mentioned twice, can you make it global?

.github/workflows/build-and-test.yml
```diff
@@ -1,33 +1,66 @@
...
+      with:
+        distribution: Ubuntu-20.04
+    - name: Download, install and run Memgraph under WSL
+      shell: wsl-bash {0} # root shell
+      run: |
+        mkdir ~/memgraph
+        curl -L https://download.memgraph.com/memgraph/v${{env.MG_VERSION}}/ubuntu-20.04/memgraph_${{env.MG_VERSION}}-community-1_amd64.deb --output ~/memgraph/memgraph-community.deb
+        dpkg -i ~/memgraph/memgraph-community.deb
+        nohup /usr/lib/memgraph/memgraph --bolt-port 7687 --data-directory="~/memgraph/data" --storage-properties-on-edges=true --storage-snapshot-interval-sec=0 --storage-wal-enabled=false --storage-recover-on-startup=false --storage-snapshot-on-exit=false --telemetry-enabled=false --log-file='' &
+        sleep 1 # Wait for Memgraph a bit.
```
jmatak, line 58: Is this simpler than running the Docker container on Windows?

jmatak (commented):
Can be approved after resolving comments :+1:

.github/workflows/build-and-test.yml
jbajic: Running docker container on windows does not work at least not on GithubActions, since by default it uses windows containres and not Linux.

.github/workflows/build-and-test.yml
jbajic: I will try it.

jbajic: `The workflow is not valid. .github/workflows/build-and-test.yml (Line: 6, Col: 21): A sequence was not expected .github/workflows/build-and-test.yml (Line: 14, Col: 25): Unrecognized named-value: 'env'. Located at position 1 within expression: env.PYTHON_VERSIONS` env variables cannot be picked up

Josipmrden (approved):

.github/workflows/build-and-test.yml
jmatak: Too bad for that, hope GitHub will add that feature for their Actions :+1:

jmatak (approved):

PR #15, opened 2021-08-18
Fix networkx transformation when ids are string

jbajic:
This PR fixes an issue when nx graph has ids which are strings.


gitbuda (approved):
Solves the problem!

PR #31, opened 2021-11-17
Add support for handling NaN values when converting nx to cypher

Josipmrden:
Added additional configuration to handle NaN values when converting from nxGraph to Cypher queries. The default configuration raises NanException. Additional handle for Nan values is provided - yielding null if nan occurs, therefore the property will not be added when injecting queries to Memgraph.

### DIFF
--- a/gqlalchemy/transformations.py
+++ b/gqlalchemy/transformations.py
@@ -21,7 +21,7 @@ import networkx as nx
 
 from gqlalchemy import Memgraph
 from gqlalchemy.models import MemgraphIndex
-from gqlalchemy.utilities import to_cypher_labels, to_cypher_properties, to_cypher_value
+from gqlalchemy.utilities import to_cypher_labels, to_cypher_properties, to_cypher_value, NetworkXCypherConfig
 
 __all__ = ("nx_to_cypher", "nx_graph_to_memgraph_parallel")
 
@@ -32,14 +32,14 @@ class NetworkXGraphConstants:
     ID = "id"
 
 
-def nx_to_cypher(graph: nx.Graph, create_index=False) -> Iterator[str]:
+def nx_to_cypher(
+    graph: nx.Graph, create_index=False, config: NetworkXCypherConfig = NetworkXCypherConfig()
+) -> Iterator[str]:
     """Generates a Cypher queries for creating graph"""
 
-    if create_index:
-        yield from _nx_nodes_to_cypher_with_index(graph)
-    else:
-        yield from _nx_nodes_to_cypher(graph)
-    yield from _nx_edges_to_cypher(graph)
+    builder = NetworkXCypherBuilder(create_index=create_index, config=config)
+
+    yield from builder.yield_queries(graph)
 
 
 def nx_graph_to_memgraph_parallel(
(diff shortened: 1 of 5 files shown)


gqlalchemy/transformations.py
```diff
@@ -32,14 +32,14 @@ class NetworkXGraphConstants:
     ID = "id"
 
 
-def nx_to_cypher(graph: nx.Graph, create_index=False) -> Iterator[str]:
+def nx_to_cypher(
+    graph: nx.Graph, create_index=False, config: NetworkXCypherConfig = NetworkXCypherConfig()
```
jmatak, line 36: Not sure how such created default named parameters would act. Not sure what is the practice if the named parameter is an object.

gqlalchemy/transformations.py
```diff
@@ -50,21 +50,21 @@ def nx_graph_to_memgraph_parallel(
...
 ) -> None:
     """Generates a Cypher queries and inserts data into Memgraph in parallel"""
-    query_groups = []
-    if create_index:
-        query_groups.append(_nx_nodes_to_cypher_with_index(graph))
-    else:
+    builder = NetworkXCypherBuilder(create_index=create_index, config=config)
+    query_groups = builder.yield_query_groups(graph)
+
+    if not create_index:
```
jmatak, line 59: Not related with this PR, but could index become part of the Builder in future releases?

jbajic: I think the index part could be moved into `NetworkXCypherConfig`

gqlalchemy/utilities.py
```diff
@@ -41,14 +59,16 @@ def to_cypher_value(value: Any) -> str:
     return f"'{value}'"
 
 
-def to_cypher_properties(properties: Optional[Dict[str, Any]] = None) -> str:
+def to_cypher_properties(
+    properties: Optional[Dict[str, Any]] = None, config: NetworkXCypherConfig = NetworkXCypherConfig()
```
jbajic, line 63: I would remove the default construction of `NetworkXCypherConfig` from all of these functions in utilities

gqlalchemy/transformations.py
jbajic: It's best to avoid default mutable arguments, I cannot see the problem here but it's generally been known to cause issues with collections (a good article that explains what could happen https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/).

jmatak (commented):
Sorry to bother this much, but I would like to express concern about how the API would look once we have Config object (meaning I was wrong the last time). After this iteration user has an option to call the API method with the config object as a calling parameter, meaning that he has no idea at all about whether that contains the index or not. 
@jbajic @Josipmrden  After seeing both, what do you think the best approach is?

Josipmrden:
> Sorry to bother this much, but I would like to express concern about how the API would look once we have Config object (meaning I was wrong the last time). After this iteration user has an option to call the API method with the config object as a calling parameter, meaning that he has no idea at all about whether that contains the index or not. @jbajic @Josipmrden After seeing both, what do you think the best approach is?

Why wouldn't he see that in the config there is an option to create the index?

jbajic (approved):
Great job! :tada:

MasterMedo:
Can this be merged?

jmatak (approved):

PR #40, opened 2021-12-07
add match support for objects

MasterMedo:
This PR adds support for querying Memgraph using Node and Relationship objects:

Here's an example
```python
from gqlalchemy import Memgraph, Node, Relationship, Match

memgraph = Memgraph("127.0.0.1", 7687)

memgraph.execute("CREATE (:Node {id: 1})-[:RELATED_TO {id: 1}]->(:Node {id: 2})")

# the first argument should be set by Memgraph
a = Node(1, ["Node"], {'id': 1})
b = Node(2, ["Node"], {'id': 2})
r = Relationship(1, "RELATED_TO", 1, 2, {'id': 1})

result = list(
    Match(memgraph.new_connection())
    .node(variable="a", node=a)
    .to(variable="r", relationship=r)
    .node(variable="b", node=b)
    .execute()
)[0]

print(result['a'])
print(result['b'])
print(result['r'])
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,19 +65,44 @@ As we can see, the example above can be error-prone, because we do not have abst
 Now, rewrite the exact same query by using the functionality of gqlalchemys query builder..
 
 ```python
-
 from gqlalchemy import Match, Memgraph
 
 memgraph = Memgraph()
 
-results = Match().node("Node",variable="from")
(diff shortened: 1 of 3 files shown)


gitbuda (approved):
Seems legit!

PR #56, opened 2021-12-13
add custom Property type class

### LINKED ISSUE #54: Create custom OGM fields instead of primitive types
Instead of using primitive types for fields of GraphObject subclasses, we can create custom fields that will make it possible to create automatic indexing and/or writing properties to disk, etc.

instead of writing:
```python
class Person(Node):
    ssn: str
    avatar: Image
```
the user will be able to write:

```python
class Person(Node):
    ssn: Property(str, index=True)
    avatar: Property(Image, on_disk=True)
```

MasterMedo:
Instead of using primitive types for fields of GraphObject subclasses, we can create custom fields that will make it possible to create automatic indexing and/or writing properties to disk, etc.

instead of writing:
```python
class Person(Node):
    ssn: str
    avatar: Image
```
the user will be able to write:

```python
class Person(Node):
    ssn: str = Field(index=True, db=Memgraph())
    avatar: Image = Field(on_disk=True, in_db=False)
```

closes #54


MasterMedo:
There is a problem that had me stumped yesterday. I'm trying to implement custom properties in a way that will make it possible to test properties, both as they are in Memgraph, and as they are in the python object representation and synchronise those.

For example, a user should be able to check if the index has been put in place in Memgraph or if a constraint is holding.
```
class Person(Node):
    name: Property(index=True, unique=True)
```

For now, I've implemented it so that upon object creation the indexes and uniqueness constraints get checked via cypher. That is a problem because if we're creating a million nodes we don't need to check every time if the constraint is in place. The other option I've considered is creating indexes on class creation. That could very well be a problem in an application that is running continuously, and the indexes get dropped. But, for now, I've ignored that. I'll create a discussion before this PR gets merged and we can implement that later as an option.

MasterMedo:
I'll put comments here how the field assigning works so it's easier to understand and write documentation for it.

```python
class Person(Node):
    name: Property(index=True, unique=True)
```

`Person` is a subclass of `Node`.
`Node` is a subclass of `GraphObject`.
`GraphObject` is a subclass of Pydantic's `BaseModel`.
`BaseModel` has a metaclass (a class that creates the `BaseModel` class).

That metaclass scans all the fields assigned to `Person`, in this case the field `name`,
and adds them to `Person.__fields__` as `FieldModel` objects.
Fields contain information such as:
- name
- type
- validators
- arguments (index, exists, unique, in_db, on_disk, ...)

MasterMedo:
Can this be merged? I want to merge it from this commit and then keep the branch alive for further improvements after the launch.

Improvements include creating proper metaclasses that will enable Index, and Constraint creation on class definition (once) instead of on every object creation. I already have some commits stashed for this and would like to push them but they are not in the working stage yet.

gqlalchemy/models.py
```diff
@@ -130,12 +131,71 @@ def __repr__(self) -> str:
         return str(self)
 
 
-class Node(UniqueGraphObject):
+class MyMeta(BaseModel.__class__):
```
gitbuda, line 134: Why `MyMeta`? Any specific reason for the name?

gqlalchemy/models.py
```diff
@@ -130,12 +131,71 @@ def __repr__(self) -> str:
...
+    def __new__(mcs, name, bases, namespace, **kwargs):  # noqa C901
+        """This creates the class `Node`. It also creates all subclasses
+        of `Node`. Whenever a class is defined as a subclass of `Node`,
+        `MyMeta.__new__` is called.
+        """
+        cls = super().__new__(mcs, name, bases, namespace, **kwargs)
+        # TODO create a discussion about accessing labels through the class definition instead of through the object. E.g. `Person.labels` instead of `person = Person("Marko"); person._node_labels`.
```
gitbuda, line 141: I think we need both because accessing labels through class definition is only possible if there is a schema. Let's discuss 😄

gqlalchemy/models.py
```diff
@@ -130,12 +131,71 @@ def __repr__(self) -> str:
...
+        cls._type = kwargs.get("_type", name)
+        cls._node_labels = kwargs.get("_node_labels", set(cls._type.split(":")))
+        labels = ":".join(cls._node_labels)
+        # TODO check if *_type* or *_node_labels* is in fields
```
gitbuda, line 145: Why not check in this PR? What should happen in that case?

tests/ogm/custom_fields_test.py
```diff
@@ -0,0 +1,62 @@
...
+    assert actual_constraints == [memgraph_constraint]
+
+
+def test_create_index(memgraph):
+    memgraph_index = MemgraphIndex("Node2", "id")
+
+    memgraph.create_index(memgraph_index)
+    actual_constraints = memgraph.get_indexes()
+
+    assert actual_constraints == [memgraph_index]
```
gitbuda, line 62: Test cases seem fine 😄

gqlalchemy/models.py
```diff
@@ -130,12 +131,71 @@ def __repr__(self) -> str:
...
+                        "Can't have an index on a property without providing"
+                        " the database `db` object.\n"
+                        "Define your property as:\n"
+                        f" {field}: type = Field(unique=True, db=Memgraph())"
+                    )
+                constraint = MemgraphConstraintUnique(labels, field)
+                db.create_constraint(constraint)
+
+            # if "on_disk" in attrs:
+            # if "use_in_db" in attrs:
```
gitbuda, line 187: These are TODOs for the subsequent PRs?

gqlalchemy/models.py
MasterMedo: Good question, should be renamed! `NodeMetaclass` sounds like a reasonable name.

gqlalchemy/models.py
MasterMedo: haha you're resolving future discussions, I love it. But yeah, I agree, both is definitely better than only one. The discussion I wanted to have is about these preceding underscores. For the class definitions we don't need the underscores because users won't set any fields/attributes to the class, but they will set attributes to the objects of the class. But then, I don't want to have different interfaces...

gqlalchemy/models.py
MasterMedo: Oh, I wanted to ask you about that. What if users don't want to create a sublcass of `Node` and they just want to use `Node` directly, this makes this much more complex (this is what that TODO would solve), I don't want to go into that on this launch.

This would enable the user to define special labels on which the index would be created, e.g.

```python
class Person(Node, _type="Person"):
    id: int = Field(index=True, _type="Person:Marko", db=Memgraph())
```

In this example, the index wouldn't be created on the label `Person` but rather on `Person:Marko`. I'm not sure how often this would be used.

tests/ogm/custom_fields_test.py
MasterMedo: Yeah, they are quite minimal, but I copied them from the other tests we had :p

gqlalchemy/models.py
MasterMedo: Exactly, they are here so I don't forget them for the next PR, which will be about adding external on_disk database support.

gqlalchemy/models.py
MasterMedo: Here is a cool example:

```python
class Person(Node):
    pass
    
class Student(Person):
    pass
```

In this example, the class Student will have only one label "Student", but a Student object might have multiple labels, e.g. "Person:Student". This is why we need both.

gqlalchemy/models.py
gitbuda: Let's move this under discussion, I wouldn't support this now.

PR #58, opened 2022-01-03
add property database interface

MasterMedo:
Tasks:

- [x] Define on-disk database interface
- [x] Implement interface for SQLite
- [x] Write unit tests
- [x] Integrate with `GraphObject`'s `on_disk` `Field` argument
- [x] Write integration tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # temp files
-on_disk_properties.db
+tests/on_disk_storage.db
 
 # Byte-compiled / optimized / DLL files
 **/__pycache__/
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 # temp files
 on_disk_properties.db
-tmp.py
 
 # Byte-compiled / optimized / DLL files
 **/__pycache__/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+# temp files
+on_disk_properties.db
+tmp.py
+
 # Byte-compiled / optimized / DLL files
 **/__pycache__/
 *.py[cod]
--- a/gqlalchemy/disk_storage.py
+++ b/gqlalchemy/disk_storage.py
@@ -15,7 +15,7 @@
 import sqlite3
 import contextlib
 
-from typing import List
+from typing import Optional, List
 
 
 class OnDiskPropertyDatabase:
@@ -80,13 +80,21 @@ class SQLitePropertyDatabase(OnDiskPropertyDatabase):
             "DO UPDATE SET property_value=excluded.property_value;"
         )
 
-    def load_node_property(self, node_id: int, property_name: str):
-        return self.execute_query(
+    def load_node_property(self, node_id: int, property_name: str) -> Optional[str]:
+        result = self.execute_query(
             "SELECT property_value "
             "FROM node_properties AS db "
             f"WHERE db.node_id = {node_id} "
             f"AND db.property_name = '{property_name}'"
         )
+
+        if len(result) == 0:
+            return None
+
+        # primary key is unique
+        assert len(result) == 1 and len(result[0]) == 1
+
+        return result[0][0]
 
     def save_relationship_property(self, relationship_id: int, property_name: str, property_value: str):
         self.execute_query(
@@ -96,10 +104,18 @@ class SQLitePropertyDatabase(OnDiskPropertyDatabase):
             "DO UPDATE SET property_value=excluded.property_value;"
         )
 
-    def load_relationship_property(self, relationship_id: int, property_name: str):
-        return self.execute_query(
+    def load_relationship_property(self, relationship_id: int, property_name: str) -> Optional[str]:
+        result = self.execute_query(
             "SELECT property_value "
             "FROM relationship_properties AS db "
             f"WHERE db.relationship_id = {relationship_id} "
             f"AND db.property_name = '{property_name}'"
         )
+
+        if len(result) == 0:
+            return None
+
+        # primary key is unique
+        assert len(result) == 1 and len(result[0]) == 1
+
+        return result[0][0]
--- a/gqlalchemy/disk_storage.py
+++ b/gqlalchemy/disk_storage.py
@@ -48,7 +48,7 @@ class SQLitePropertyDatabase(OnDiskPropertyDatabase):
                     cursor.execute(query)
                     return cursor.fetchall()
 
-    def create_node_property_table(self):
+    def create_node_property_table(self) -> None:
         self.execute_query(
             "CREATE TABLE IF NOT EXISTS node_properties ("
             "node_id integer NOT NULL,"
--- a/gqlalchemy/disk_storage.py
+++ b/gqlalchemy/disk_storage.py
@@ -15,23 +15,30 @@
 import sqlite3
 import contextlib
 
+from abc import ABC
 from typing import Optional, List
 
 
-class OnDiskPropertyDatabase:
-    def save_node_property(self, node_id: int, property_name: str, property_value: str):
+class OnDiskPropertyDatabase(ABC):
+    def save_node_property(self, node_id: int, property_name: str, property_value: str) -> None:
         pass
 
-    def load_node_property(self, node_id: int, property_name: str, property_value: str):
+    def load_node_property(self, node_id: int, property_name: str, property_value: str) -> Optional[str]:
         pass
 
-    def save_relationship_property(self, relationship_id: int, property_name: str, property_value: str):
+    def delete_node_property(self, node_id: int, property_name: str, property_value: str) -> None:
         pass
 
-    def load_relationship_property(self, relationship_id: int, property_name: str, property_value: str):
+    def save_relationship_property(self, relationship_id: int, property_name: str, property_value: str) -> None:
         pass
 
-    def drop_database(self):
+    def load_relationship_property(self, relationship_id: int, property_name: str, property_value: str) -> Opitional[str]:
+        pass
+
+    def delete_relationship_property(self, node_id: int, property_name: str, property_value: str) -> None:
+        pass
+
+    def drop_database(self) -> None:
         pass
 
 
@@ -41,7 +48,7 @@ class SQLitePropertyDatabase(OnDiskPropertyDatabase):
         self.create_node_property_table()
         self.create_relationship_property_table()
 
-    def execute_query(self, query) -> List[str]:
+    def execute_query(self, query: str) -> List[str]:
         with contextlib.closing(sqlite3.connect(self.database_name)) as conn:
             with conn:  # autocommit changes
                 with contextlib.closing(conn.cursor()) as cursor:
@@ -58,7 +65,7 @@ class SQLitePropertyDatabase(OnDiskPropertyDatabase):
             ");"
         )
 
-    def create_relationship_property_table(self):
+    def create_relationship_property_table(self) -> None:
         self.execute_query(
             "CREATE TABLE IF NOT EXISTS relationship_properties ("
             "relationship_id integer NOT NULL,"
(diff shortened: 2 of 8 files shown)


gqlalchemy/disk_storage.py
```diff
@@ -0,0 +1,121 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sqlite3
+import contextlib
+
+from typing import Optional, List
+
+
+class OnDiskPropertyDatabase:
```
gitbuda, line 21: Use https://docs.python.org/3/library/abc.html here?

gqlalchemy/disk_storage.py
```diff
@@ -0,0 +1,121 @@
...
+    def load_node_property(self, node_id: int, property_name: str, property_value: str):
+        pass
+
+    def save_relationship_property(self, relationship_id: int, property_name: str, property_value: str):
+        pass
+
+    def load_relationship_property(self, relationship_id: int, property_name: str, property_value: str):
+        pass
+
+    def drop_database(self):
```
gitbuda, line 34: Add `delete_xyz_property` since these properties could be big?

.gitignore
```diff
@@ -1,3 +1,7 @@
+# temp files
+on_disk_properties.db
+tmp.py
```
jbajic, line 3: These seem very specific for what you are doing, I suggest removing and using global git ignore(https://stackoverflow.com/questions/7335420/global-git-ignore)

gqlalchemy/disk_storage.py
```diff
@@ -0,0 +1,121 @@
...
+        pass
+
+
+class SQLitePropertyDatabase(OnDiskPropertyDatabase):
+    def __init__(self, database_name: str = "on_disk_properties.db"):
+        self.database_name = database_name
+        self.create_node_property_table()
+        self.create_relationship_property_table()
+
+    def execute_query(self, query) -> List[str]:
```
jbajic, line 44: query typing

gqlalchemy/disk_storage.py
```diff
@@ -0,0 +1,121 @@
...
+        with contextlib.closing(sqlite3.connect(self.database_name)) as conn:
+            with conn:  # autocommit changes
+                with contextlib.closing(conn.cursor()) as cursor:
+                    cursor.execute(query)
+                    return cursor.fetchall()
+
+    def create_node_property_table(self):
```
jbajic, line 51: if nothing is returned I suggest annotating it
```suggestion
    def create_node_property_table(self) -> None:
```

.gitignore
MasterMedo: `on_disk_properties.db` is used by tests. I can move it to `/test/`. I agree regarding the `tmp.py`.

gqlalchemy/disk_storage.py
MasterMedo: good catch.

jbajic (commented):
Also update the copyright year for every file you change

MasterMedo:
I think I'll open a separate PR for that ^^

tests/on_disk_property_storage/test_multiprocess.py
```diff
@@ -0,0 +1,79 @@
...
+
+import pytest
+import random
+
+from gqlalchemy import Memgraph, Node, Field, SQLitePropertyDatabase
+
+
+db = Memgraph()
+on_disk_db = SQLitePropertyDatabase("./tests/on_disk_storage.db")
+db.add_on_disk_storage(on_disk_db)
```
gitbuda, line 26: `add_on_disk_storage` seems to be a bit misleading (at least to me), what about `init_disk_storage`?

gqlalchemy/memgraph.py
```diff
@@ -318,12 +364,22 @@ def load_relationship_with_start_node_and_end_node(
...
         else:
             raise GQLAlchemyError("Can't create a relationship without start_node_id and end_node_id.")
 
+        if self.on_disk_db:
+            for field in relationship.__fields__:
+                value = getattr(relationship, field, None)
+                if value is not None:
+                    if "on_disk" in relationship.__fields__[field].field_info.extra:
+                        self.on_disk_db.save_relationship_property(result._id, field, value)
+                        setattr(result, field, value)
```
gitbuda, line 379: What if `on_disk_db` is `None` and `on_disk` field is `True`, will that raise an error?

tests/on_disk_property_storage/test_multiprocess.py
```diff
@@ -0,0 +1,79 @@
...
+    number_of_processes = mp.cpu_count() // 2 + 1
+    chunk_size = n // number_of_processes
+    processes = []
+    for i in range(number_of_processes):
+        processes.append(mp.Process(target=_create_n_user_objects, args=(chunk_size,)))
+
+    for p in processes:
+        p.start()
+    for p in processes:
+        p.join()
```
gitbuda, line 66: A bit better here would be to use `Pool` https://docs.python.org/3/library/multiprocessing.html#using-a-pool-of-workers because you can make the same test run on a pool of threads and a pool of processes + the code should be shorter 😄

tests/on_disk_property_storage/test_multiprocess.py
MasterMedo: Sure, that sounds good. I was also thinking of
```python
db = Memgraph()
on_disk_db = SQLitePropertyDatabase("./tests/on_disk_storage.db")
on_disk_db.app_init(db)
```
That is used usually in flask based services. What do you think about that?

gqlalchemy/memgraph.py
MasterMedo: if `on_disk_db` is None, that will fail the first if on line 373. But I can make it so it says `if self.on_disk_db is not None`, that is better practice. Or did you have something else in mind?

tests/on_disk_property_storage/test_multiprocess.py
MasterMedo: Sure, that sounds good ^^
I got used to this syntax back in python 2.7 :sweat_smile:

MasterMedo:
```suggestion
    number_of_processes = mp.cpu_count() // 2 + 1
    chunk_size = n // number_of_processes
    with mp.Pool(processes=number_of_processes) as pool:
        pool.map(_create_n_user_objects, [chunk_size] * number_of_processes)
```

gqlalchemy/disk_storage.py
```diff
@@ -0,0 +1,146 @@
...
+        self.create_relationship_property_table()
+
+    def execute_query(self, query: str) -> List[str]:
+        with contextlib.closing(sqlite3.connect(self.database_name)) as conn:
+            with conn:  # autocommit changes
+                with contextlib.closing(conn.cursor()) as cursor:
+                    cursor.execute(query)
+                    return cursor.fetchall()
+
+    def create_node_property_table(self) -> None:
```
jbajic, line 60: I assume they were meant to be private since they are being called in `__init__`
```suggestion
    def _execute_query(self, query: str) -> List[str]:
        with contextlib.closing(sqlite3.connect(self.database_name)) as conn:
            with conn:  # autocommit changes
                with contextlib.closing(conn.cursor()) as cursor:
                    cursor.execute(query)
                    return cursor.fetchall()

    def _create_node_property_table(self) -> None:
```

gqlalchemy/memgraph.py
MasterMedo:
```suggestion
        if self.on_disk_db is not None:
            for field in relationship.__fields__:
                value = getattr(relationship, field, None)
                if value is not None:
                    if "on_disk" in relationship.__fields__[field].field_info.extra:
                        self.on_disk_db.save_relationship_property(result._id, field, value)
                        setattr(result, field, value)
```

gqlalchemy/memgraph.py
```diff
@@ -236,11 +246,20 @@ def save_node(self, node: Node):
...
-            return self.create_node(node)
+            result = self.create_node(node)
+
+        if self.on_disk_db:
+            for field in node.__fields__:
+                value = getattr(node, field, None)
+                if value is not None:
+                    if "on_disk" in node.__fields__[field].field_info.extra:
+                        self.on_disk_db.save_node_property(result._id, field, value)
+                        setattr(result, field, value)
```
jbajic, line 261: Maybe split this into new method => `_save_property_in_disk_db`

gqlalchemy/memgraph.py
```diff
@@ -254,14 +273,26 @@ def save_node_with_id(self, node: Node) -> Optional[Node]:
...
+
+        if self.on_disk_db:
+            for field in result.__fields__:
+                value = getattr(result, field, None)
+                if "on_disk" in result.__fields__[field].field_info.extra:
+                    try:
+                        new_value = self.on_disk_db.load_node_property(result._id, field)
+                    except sqlite3.OperationalError:
+                        new_value = value
+                    setattr(result, field, new_value)
```
jbajic, line 293: Also maybe split

MasterMedo: yep, that's a good idea.

gqlalchemy/memgraph.py
MasterMedo: yep, will do.

gqlalchemy/disk_storage.py
MasterMedo: Ammm, create_node_property_table can be private, sure, but there's no harm in exposing it? Execute query should definitely be exposed.

tests/on_disk_property_storage/test_multiprocess.py
gitbuda: Why not just `init`?

gitbuda: Or, just in the constructor?

gqlalchemy/memgraph.py
gitbuda: I think I had something else in mind, what if there is an on_disk prop, but on_disk storage is not defined? That has to fail because it's not good if something the user wanted to be on_disk ends up in-memory. In other words, is it possible to miss-use this API and get inconsistent behavior? NOTE: I didn't check the code deeply. I'm just reasoning based on the small part of the code here.

tests/on_disk_property_storage/test_multiprocess.py
MasterMedo: Yeah, Sqlalchemy provides both possibilities:
```python
Sqlalchemy(app)
```
and
```python
db = Sqlalchemy()
db.init_app(app)
```

I don't know why not just `init`, it's a standard with flask based services.
https://stackoverflow.com/questions/58042067/what-is-purpose-of-init-app-function-in-flask

gqlalchemy/memgraph.py
MasterMedo: Oh, I think I understand what you're asking. There is undefined behaviour, yes. If the user specifies the property that should go on disk, but he didn't provide an `on_disk_db`, those properties would get dropped. We can throw a `GQLAlchemyError` in those instances upon class init, e.g.:
```
db = Memgraph()

class User(Node):
    huge_string: str = Field(on_disk_db=True, db=db)
```

This should throw an error:
```python
`db` object in line 4, file `tmp.py` doesn't have `on_disk_db` specified.
    huge_string: str = Field(on_disk_db=True, db=db)
    
Add an on_disk_db like this:
from gqlalchemy import Memgraph, SQLOnDiskStorage
db = Memgraph()
SQLOnDiskStorage(db)
```

tests/on_disk_property_storage/test_multiprocess.py
MasterMedo: I've added the `__init__` approach and renamed the function to `init_disk_storage`

PR #61, opened 2022-01-04
OGM: saving python objects to Memgraph

MasterMedo:
This PR aims to create a way for Python objects to be saved into Memgraph using Merge and Create. The operations have to support only internal `Field` objects and respect the arguments given to them (`index`, `exists`, `unique`, primary_key`, `on_disk`)...

related:

#60 

- [x] 1. simple create (no field arguments, node was never in Memgraph)
- [x] 2. save node to Memgraph after it was at least once saved (no field arguments)
- [x] 3. save node when the node has unique fields
- [ ] ~~4. save node when the node has edges~~
- [ ] ~~5. save node when the node has index fields~~

# 1. simple create (no field arguments, node was never in Memgraph)
Here is a list of all supported data types.
The node gets **created** in Memgraph, `MATCH` and `MERGE` are not used in the cypher, only `CREATE` is used.
```python
class MyNode(Node):
    name: Optional[str] = Field()
    done: Optional[bool]
    unit: Optional[int] = Field()
    time: Optional[float]
    keys: Optional[list[int]]
    tags: Optional[dict[str, float]]
    date: Optional[datetime.datetime] = Field()


# initialise python object
node = MyNode(name="First")
# creates the node in the GraphDatabase
node.save(db)
```

# 2. save node to Memgraph after it was at least once saved (no field arguments)
When the node is first inserted into Memgraph, the Memgraph internal id is set on the node object in `node._id` field. The internal id is from then on used to `MERGE` the node on that internal id. The properties are `SET` after the node is fetched or created (merged).
```python
class MyNode(Node):
    name: Optional[str] = Field()

node = MyNode(name="First")
node.save(db)  # creates a node in Memgraph and fetches the internal Memgraph node id to the node._id
node.name = "Second"
node.save(db)  # fetches the node using the internal id and then updates the properties
```

# 3. save node when the node has unique fields
- use all non null unique fields to get all nodes that have either unique field set to the current value
- if more than 1 node is found in Memgraph, a conflict happend and the node will not be inserted (An exception is thrown)

# 4. save node when the node has edges
- what if node has no edges in Memgraph
- what if edges in Memgraph changed
**CONCLUSION**: not use edges for the first release

# 5. save node when the node has index fields
- what if indexed field has not yet been submitted in Memgraph
- what if indexed field was changed in Memgraph
**CONCLUSION**: not use indexes for the first release

### DIFF
--- a/gqlalchemy/memgraph.py
+++ b/gqlalchemy/memgraph.py
@@ -169,7 +169,9 @@ class Memgraph:
 
     def _get_nodes_with_unique_fields(self, node: Node) -> Optional[Node]:
         return self.execute_and_fetch(
-            f"MATCH (node: {node._label}) WHERE {node._get_cypher_unique_fields_or_block('node')} RETURN node;"
+            f"MATCH (node: {node._label})"
+            + f" WHERE {node._get_cypher_unique_fields_or_block('node')}"
+            + " RETURN node;"
         )
 
     def get_variable_assume_one(self, query_result: Iterator[Dict[str, Any]], variable_name: str) -> Any:
@@ -208,7 +210,10 @@ class Memgraph:
 
     def save_node_with_id(self, node: Node) -> Optional[Node]:
         results = self.execute_and_fetch(
-            f"MATCH (node: {node._label}) WHERE id(node) = {node._id} {node._get_cypher_set_properties('node')} RETURN node;"
+            f"MATCH (node: {node._label})"
+            + f" WHERE id(node) = {node._id}"
+            + f" {node._get_cypher_set_properties('node')}"
+            + " RETURN node;"
         )
 
         return self.get_variable_assume_one(results, "node")
@@ -226,7 +231,7 @@ class Memgraph:
 
     def load_node_with_all_properties(self, node: Node) -> Optional[Node]:
         results = self.execute_and_fetch(
-            f"MATCH (node: {node._label} WHERE {node._get_cypher_fields_and_block('node')} RETURN node;"
+            f"MATCH (node: {node._label}) WHERE {node._get_cypher_fields_and_block('node')} RETURN node;"
         )
         return self.get_variable_assume_one(results, "node")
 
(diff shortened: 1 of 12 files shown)


gqlalchemy/models.py
```diff
@@ -131,94 +165,131 @@ def __repr__(self) -> str:
...
+            if "unique" in self.__fields__[field].field_info.extra:
+                if getattr(self, field) is not None:
+                    return True
+        return False
+
+    @property
+    def _label(self) -> str:
+        return ":".join(sorted(self._labels))
+
+    def save(self, db: "Memgraph") -> None:  # noqa F821
```
MasterMedo, line 255:
```suggestion
    def save(self, db: "Memgraph") -> Node:  # noqa F821
```

gqlalchemy/models.py
```diff
@@ -131,94 +165,131 @@ def __repr__(self) -> str:
...
+        node = db.save_node(self)
+        for field in self.__fields__:
+            setattr(self, field, getattr(node, field))
+        self._id = node._id
+        return self
+
+    def load(self, db: "Memgraph") -> None:  # noqa F821
```
MasterMedo, line 262:
```suggestion
    def load(self, db: "Memgraph") -> Node:  # noqa F821
```

gqlalchemy/models.py
```diff
@@ -229,13 +300,29 @@ def __str__(self) -> str:
...
         )
 
+    def save(self, db: "Memgraph") -> None:  # noqa F821
+        relationship = db.save_relationship(self)
+        for field in self.__fields__:
+            setattr(self, field, getattr(relationship, field))
+        self._id = relationship._id
+        return self
+
+    def load(self, db: "Memgraph") -> None:  # noqa F821
```
MasterMedo, line 319:
```suggestion
    def load(self, db: "Memgraph") -> Relationship:  # noqa F821
```

gqlalchemy/memgraph.py
```diff
@@ -144,3 +166,147 @@ def new_connection(self) -> Connection:
...
+        return self.get_variable_assume_one(results, "relationship")
+
+    def load_relationship_with_all_properties(self, relationship: Relationship) -> Optional[Relationship]:
+        where_block = relationship._get_cypher_fields_or_block("relationship")
+        results = self.execute_and_fetch(
+            f"MATCH ()-[relationship: {relationship._type}]->() WHERE {where_block} RETURN relationship"
+        )
+        return self.get_variable_assume_one(results, "relationship")
+
+    def save_relationship(self, relationship: Relationship) -> Optional[Relationship]:
```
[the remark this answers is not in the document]
MasterMedo: This can be avoided by not being able to update relationshis without internal id. Before updating a relationship you first have to create it or load it from memgraph.

PR #66, opened 2022-01-12
Add create stream support

gitbuda:


gqlalchemy/models.py
```diff
@@ -60,6 +60,57 @@ def to_cypher(self) -> str:
...
+        query = f"CREATE KAFKA STREAM {self.name} TOPICS {topics} TRANSFORM {self.transform} "
+        if self.consumer_group is not None:
+            query += f"CONSUMER_GROUP {self.consumer_group} "
+        if self.batch_interval is not None:
+            query += f"BATCH_INTERVAL {self.batch_interval} "
+        if self.batch_size is not None:
+            query += f"BATCH_SIZE {self.batch_size} "
+        if self.bootstrap_servers is not None:
+            query += f"BOOTSTRAP_SERVERS {self.bootstrap_servers} "
+        return query
```
g-despot, line 92: Add ,

MasterMedo:
Hm, end to end tests should be added, but ok, we can do that after the merge in a future pull request.

tests/intergration/test_stream.py
```diff
@@ -0,0 +1,89 @@
...
+    kafka_stream = MemgraphKafkaStream(
+        name="test_stream",
+        topics=["topic"],
+        transform="kafka_stream.transform",
+        consumer_group="my_group",
+        batch_interval="9999",
+        bootstrap_servers="'localhost:9092'",
+    )
+    query = "CREATE KAFKA STREAM test_stream TOPICS topic TRANSFORM kafka_stream.transform CONSUMER_GROUP my_group BATCH_INTERVAL 9999 BOOTSTRAP_SERVERS 'localhost:9092';"
+    assert kafka_stream.to_cypher() == query
```
gitbuda, line 76: @g-despot try also to execute this query because that will cover most (actually the whole implementation) of `create_stream`, `get_streams`, and `drop_stream`

tests/intergration/test_stream.py
```diff
@@ -0,0 +1,89 @@
+# Copyright (c) 2016-2021 Memgraph Ltd. [https://memgraph.com]
```
gitbuda, line 1: Update to 2022 because that's s new file :D (the same applies to the other PRs)

tests/intergration/test_stream.py
gitbuda: Aha, that's tested before 😄

MasterMedo: The `get_streams` and `drop_stream` are functions in `Memgraph` that don't have a state, so you can't easily inspect the cypher they create. And it is not a reliable test if the implementation ever changes. These stream tests do not actually create any streams in Memgraph because mgclient throws:

```
mgclient.DatabaseError: Failed to initialize Kafka consumer test_stream : Local: Broker transport failure
```

I can add the test declarations so that future implementations don't forget to cover these cases.

tests/intergration/test_stream.py
MasterMedo: We have to update many files, I think it's easier to do it once all PR's are merged in an "bump year" PR

tests/intergration/test_stream.py
```diff
@@ -0,0 +1,89 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import pytest
+from gqlalchemy import MemgraphKafkaStream, MemgraphPulsarStream
+
+
+def stream_exists(streams, stream_name):
+    return any(map(lambda s: s["name"] == stream_name, streams))
```
MasterMedo, line 19:
```suggestion
def stream_exists(stream: str, memgraph: Memgraph) -> bool:
    return stream in memgraph.get_streams()
```

MasterMedo (approved):
LGTM

PR #70, opened 2022-01-12
Add new Cypher clauses to the query builder

g-despot:
Add Cypher clauses to the query builder:

- [x] CALL
- [x] WITH
- [x] UNWIND
- [x] RETURN
- [x] YIELD
- [x] ORDER BY
- [x] LIMIT
- [x] CREATE
- [x] MERGE


gitbuda:
@g-despot when the code is done (decision about how exactly the object will be called), `README.md` also has to be updated 😄

g-despot:
@MasterMedo check out this PR when you get the chance so I know if I am going in the right direction with implementing all the missing Cypher queries in the query builder. Still missing: `DELETE`, `LOAD CSV`, `MERGE`, `REMOVE`, `SET` and `UNION`.
I am currently missing an error handling layer and a logic layer to check if the constructed query is following basic Cypher syntax rules.

MasterMedo:
I gotta admit, I greatly dislike the name `G`. I think it's better to be explicit. Class names should be self-explanatory. I'd much rather if we'd create a `DeclarativeBase` class and exposed it's starting properties as objects to be imported e.g.
```python
match = DeclarativeBase().match
merge = DeclarativeBase().merge
```
.
```python
from gqlalchemy import match, merge
```

Another idea is to just rename the `G` object to `QueryBuilder` or some other better name.

gqlalchemy/query_builder.py
```diff
@@ -146,24 +168,123 @@ def construct_query(self) -> str:
...
+
+        self._results = results
+
+    @property
+    def results(self) -> str:
+        return self._results if self._results is not None else ""
+
+    def construct_query(self) -> str:
+        if len(self.results) == 0:
+            return f" WITH * "
```
MasterMedo, line 208: when does this happen?

gqlalchemy/query_builder.py
```diff
@@ -146,24 +168,123 @@ def construct_query(self) -> str:
...
+    def __init__(self, limit: int):
+        super().__init__(GTypes.LIMIT)
+
+        self.limit = limit
+
+    def construct_query(self) -> str:
+        return f" LIMIT {self.limit} "
+
+
+class G:
```
MasterMedo, line 265: Suggested names: `QueryBuilder`, `DeclarativeBase`, `QueryBase`

MasterMedo (changes requested):

gqlalchemy/query_builder.py
```diff
@@ -146,24 +168,163 @@ def construct_query(self) -> str:
...
+        self.list_expression = list_expression
+        self.variable = variable
+
+    def construct_query(self) -> str:
+        return f" UNWIND {self.list_expression} AS {self.variable} "
+
+
+def dict_to_alias_statement(alias_dict: Dict[str, str]) -> str:
+    return ", ".join(
+        f"{key} AS {value}" if value != "" and key != value else f"{key}" for (key, value) in alias_dict.items()
```
gitbuda, line 184: @g-despot great improvement!

MasterMedo (changes requested):
Create and Merge need to be added.

MasterMedo (approved):

PR #75, opened 2022-01-19
Add Kafka and Pulsar consumer

g-despot:
Add a Kafka and Pulsar consumer to GQLAlchemy. This functionality is helpful if Memgraph pushes messages to a cluster and users want to communicate with Memgraph via Kafa/Pulsar instead of Bolt. 

This PR probably won't be ready for the GQLAlchemy 1.1 release. The consumers should be extended into producers and a stable testing framework is needed. The tests are currently dependant on an unstable Kafka and Pulsar cluster.

### DIFF
--- a/gqlalchemy/stream.py
+++ b/gqlalchemy/stream.py
@@ -1,0 +1,56 @@
+from kafka import KafkaConsumer as KafkaConsumer_
+from time import sleep
+from abc import ABC, abstractmethod
+import pulsar
+import threading
+
+
+class Consumer(ABC):
+    @abstractmethod
+    def _consume_with_function(self):
+        pass
+
+    def consume_with_function(self, function, sleep_duration):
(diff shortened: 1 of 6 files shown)


gqlalchemy/stream.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class PulsarConsumer(Consumer):
+    def __init__(self, *args, **kwargs):
+        self.client = pulsar.Client(*args, **kwargs)
+
+    def subscribe(self, *args, **kwargs):
+        self.consumer = self.client.subscribe(*args, **kwargs)
+
+    def _consume_with_function(self, function, sleep_duration):
```
gitbuda, line 46: `sleep_duration` should have a default value + consider renaming `consume_with_function` to just `consume`

PR #76, opened 2022-01-19
Add custom string method to query builder

g-despot:
The `add_custom_cypher()` method enables the user to add any string to the query builder. It is needed because there are functionalities that still need to be implemented inside the query builder.


gitbuda:
@g-despot, could you put a simple description of what were/are the reasons to add this. I understand what's the PR about, but I'm interested what was the exact problem 😄

MasterMedo:
I think the idea is to enable the user to write any custom queries that are not supported yet. Maybe the name of the function should be `add_custom_cypher`.

Issue #79, opened 2022-01-20
Load() doesn't work without default or Optional


katarinasupe:
It would be great that the `load()` method works if we add only one property upon which object in Memgraph is found. 

For example, let's say we have defined Stream class like this:
```
class Stream(Node, labels={"User", "Stream"}):
    name: str = Field(index=True, unique=True, db=memgraph, label="User")
    id: str = Field(index=True, unique=True, db=memgraph)
    url: str = Field()
    followers: int = Field()
    createdAt: str = Field()
    totalViewCount: int = Field()
    description: str = Field()
```
Now, let's suppose I want to find a streamer based on its `id` (which is unique). I expect that `stream = Stream(id="7").load(db=memgraph)` returns `Stream` whose `id` is "7". That does not work, because it's expected that I give value for all properties of class `Stream`. This can be solved by adding default value ("") to each property, or by using Optional for now.

MasterMedo:
Hm, after giving it some thought, we don't need to create anything. The class `Node` should work for this purpose.

```python
stream = memgraph.load_node(Node(
    id="7",
    _label="Stream"
))
```

This will return an instance of `Stream`. If you were to do a `Node(...).load(db=memgraph)` you would have gotten an instance of `Node`.

I think this is plenty good for this use-case. I would close this issue if you agree.

PR #86, opened 2022-01-26
Add load node test

katarinasupe:
I added a test related to closed issue https://github.com/memgraph/gqlalchemy/issues/79.


MasterMedo:
You mentioned the solution is not working for you, but these tests show that everything was executed correctly. The reason the tests are failing is because the files aren't formatted correctly. The linter and formatter fail the tests.

You should run flake8 and black before pushing (or use a draft PR).

katarinasupe:
I have run flake8 and black, but locally I get this error:
`
FAILED tests/ogm/load_node_test.py::test_load_node - mgclient.DatabaseError: line 1:36 extraneous input 'node' expecting {<EOF>, ';'}`

MasterMedo:
> I have run flake8 and black, but locally I get this error: ` FAILED tests/ogm/load_node_test.py::test_load_node - mgclient.DatabaseError: line 1:36 extraneous input 'node' expecting {<EOF>, ';'}`

You ran black and flake8 with poetry?
Did you run `poetry install` before that?
What version of pymgclient are you using?

katarinasupe:
First I run poetry install, then flake8 and black with poetry. I don't have any problems with testing on other PR (https://github.com/memgraph/docs/pull/292), but here I keep getting the same error. I am using pymgclient 1.1.0.

katarinasupe:
I added a test where I defined property without Field, and it's still being saved in Memgraph.

MasterMedo:
I think this test shows a flaw with the way tsome tests were defined. Can you try renaming the test file to `test_load_node`? I think all test files should be starting with `test`. I probably broke that rule at some point but didn't notice it up until now because I always manually tested the files individually until they worked.

katarinasupe:
`test_without_field` fails, as expected. 

We thought that if we define a class like this:

```
 class User(Node):
        name: str = Field(index=True, unique=True, db=memgraph)
        followers: int
```
        
that the `followers` property won't be saved into the database. This is wrong since `followers` property is saved in the database.

katarinasupe:
Removed `test_without_field`, since we still don't have that feature. I'll add the test after implementing the feature in another PR.

g-despot (approved):

PR #91, opened 2022-02-14
[develop < T0001-GA] Memgraph instance runner

g-despot:
- [x] Implement Docker instance runner
- [x] Implement Binary instance runner
- [x] Create tests for the instance runner


gqlalchemy/instance_runner.py
```diff
@@ -0,0 +1,228 @@
...
+        Attributes:
+            restart: A bool indicating if the instance should be
+              restarted if it's already running.
+        """
+        if not restart and self.is_running():
+            return
+        self.stop()
+        self._container = self._client.containers.run(
+            image=self.docker_image.value + ":" + self.docker_image_tag,
+            command="/usr/lib/memgraph/memgraph " + (" ").join([f"{k}={v}" for k, v in self.config.items()]),
```
BorisTasevski, line 205: I would suggest promoting all strings to string constants

BorisTasevski (changes requested):

gqlalchemy/instance_runner.py
```diff
@@ -0,0 +1,228 @@
...
+        connections after the timeout period has passed.
+    """
+    start_time = time.perf_counter()
+    time.sleep(delay)
+    while True:
+        try:
+            with socket.create_connection((host, port), timeout=timeout):
+                break
+        except OSError as ex:
+            time.sleep(delay)
```
Josipmrden, line 52: We sleep 10ms every time? Is that a lot of pinging to Memgraph connection?
What do you think @gitbuda

gqlalchemy/instance_runner.py
```diff
@@ -0,0 +1,228 @@
...
+        delay: A float that defines how long to wait between retries.
+        timeout: A float that defines how long to wait for the status.
+
+    Raises:
+      TimeoutError: Raises an error when the container isn't running after the
+        timeout period has passed.
+    """
+    start_time = time.perf_counter()
+    time.sleep(delay)
+    container.reload()
```
Josipmrden, line 73: Here as well @gitbuda

gqlalchemy/instance_runner.py
```diff
@@ -0,0 +1,228 @@
...
+from .memgraph import Memgraph
+
+
+class DockerImage(Enum):
+    MEMGRAPH = "memgraph/memgraph"
+    MAGE = "memgraph/memgraph-mage"
+
+
+def wait_for_port(host: str = "127.0.0.1", port: int = 7687, delay: float = 0.01, timeout: float = 5.0) -> None:
+    """Wait for a TCP port to become available.
```
Josipmrden, line 33: What's the difference between 0.0.0.0 (below instance method) and 127.0.0.1

gqlalchemy/instance_runner.py
```diff
@@ -0,0 +1,153 @@
...
+        self.stop()
+
+        self.container = self.client.containers.run(
+            self.docker_image.value + ":" + self.docker_tag + (" ").join([f"{k} {v}" for k, v in self.config]),
+            detach=True,
+            ports={f"7687/tcp": self.port},
+        )
+        wait_for_port(self.host, self.port)
+
+        self.connect()
```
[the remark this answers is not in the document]
g-despot: I don't really see a better module for this function than the `instance_runner`.  
I could add it to the `utilities` module, but as far as I know, it's better to keep it here as the functionality is intended for checking running Memgraph instances.  Any suggestions? @BorisTasevski

gqlalchemy/instance_runner.py
g-despot: The paths and error messages have been replaced with constants.

gqlalchemy/instance_runner.py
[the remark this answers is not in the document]
BorisTasevski: What Buda proposes is that `connect()` call is removed from `start()` method, because it is counter-intuitive and does not comply with single responsibility principle. If the method is named `start`, it should only start, not connect also. Add different(i.e. `start_and_connect())` method for this

gqlalchemy/instance_runner.py
```diff
@@ -0,0 +1,233 @@
...
+DOCKER_TIMEOUT_ETTOT_MESSAGE = "Waited too long for the Docker container to start."
+MEMGRAPH_CONNECTION_ERROR_MESSAGE = "The Memgraph process probably died."
+
+
+class DockerImage(Enum):
+    MEMGRAPH = "memgraph/memgraph"
+    MAGE = "memgraph/memgraph-mage"
+
+
+def wait_for_port(host: str = "127.0.0.1", port: int = 7687, delay: float = 0.01, timeout: float = 5.0) -> None:
```
BorisTasevski, line 39: All strings should be promoted to constants, including "127.0.0.1"

gqlalchemy/instance_runner.py
```diff
@@ -0,0 +1,233 @@
...
+        timeout: A float that defines how long to wait for the status.
+
+    Raises:
+      TimeoutError: Raises an error when the container isn't running after the
+        timeout period has passed.
+    """
+    start_time = time.perf_counter()
+    time.sleep(delay)
+    container.reload()
+    while container.status != "running":
```
BorisTasevski, line 79: String -> String Constant

BorisTasevski (changes requested):

gqlalchemy/instance_runner.py
g-despot: Fixed

gqlalchemy/instance_runner.py
g-despot: Fixed

BorisTasevski (changes requested):

gqlalchemy/instance_runner.py
```diff
@@ -0,0 +1,282 @@
...
+        Attributes:
+            restart: A bool indicating if the instance should be
+              restarted if it's already running.
+        """
+        if not restart and self.is_running():
+            return
+
+        self.stop()
+        self._container = self._client.containers.run(
+            image=self.docker_image.value + ":" + self.docker_image_tag,
```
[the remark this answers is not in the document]
g-despot: You are right, I replaced two occurrences of string additions with f strings.

BorisTasevski (approved):

gqlalchemy/instance_runner.py
gitbuda: Since delay is configurable, I think it's fine, but ofc, exponential backoff would be better here 😄

gqlalchemy/instance_runner.py
gitbuda: The same answer as above 😛😂

g-despot:
I need one more review @Josipmrden, @brunos252 and @katarinasupe.

Josipmrden (commented):
Everything is good, but try to implement exponential back-off like Buda said in the unresolved comments. You increase the delay by doubling it every time it fails, until it succeeds.

gqlalchemy/instance_runner.py
```diff
@@ -0,0 +1,287 @@
...
+    while True:
+        try:
+            with socket.create_connection((host, port), timeout=timeout):
+                break
+        except OSError as ex:
+            time.sleep(delay * 2**retries)
+            if time.perf_counter() - start_time >= timeout:
+                raise TimeoutError(TIMEOUT_ERROR_MESSAGE.format(port=port, host=host)) from ex
+
+        retries += 1
```
Josipmrden, line 79: Multiplication by 2 is a much less costly operation than power by 2, especially with floats.
Consider doing something like:

```
backoff = delay
while True:
...
time.sleep(backoff)
backoff *= 2
```

Josipmrden (commented):
Comment again, but just stilistically :), looks good now!

katarinasupe (approved):

PR #92, opened 2022-02-14
[develop < T0005-GA] Add execute option to query builder

g-despot:
- [x] Save a list of clauses that are used when constructing queries with the query builder
- [x] Check if the `RETURN` clause was used in the query builder to call `execute()` instead of `execute_and_fetch()`


gqlalchemy/query_builder.py
```diff
@@ -372,28 +372,33 @@ class DeclarativeBase(ABC):
     def __init__(self, connection: Optional[Union[Connection, Memgraph]] = None):
         self._query: List[Any] = []
         self._connection = connection if connection is not None else Memgraph()
+        self._clauses: List[DeclarativeBaseTypes] = []
```
BorisTasevski, line 375: Are we planning some extensive use of this list of clauses? It seems to me that as of now it is being read from only to check if there is "RETURN" statement. If we have some plans around this list, then ok, otherwise, i would replace it with simple boolean

BorisTasevski (commented):
self._clauses in DeclarativeBase seem overengineered as of now. Let's iterate on it a bit

Josipmrden (approved):

BorisTasevski (approved):

PR #95, opened 2022-02-23
Feature/migrate implementation from demo

BorisTasevski:
In this PR You can see the functionality developed for demo, that is identified as suitable to be migrated to gql alchemy.

### DIFF
--- a/gqlalchemy/loaders.py
+++ b/gqlalchemy/loaders.py
@@ -1,0 +1,386 @@
+# Copyright (c) 2016-2022 Memgraph Ltd. [https://memgraph.com]
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from . import Memgraph
+from .query_builder import QueryBuilder, Unwind
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from gqlalchemy.models import (
+    MemgraphIndex, 
+    MemgraphTrigger, 
+    TriggerEventObject, 
+    TriggerEventType, 
+    TriggerExecutionPhase
+)
+from pyarrow import fs
+from typing import (
+    List, 
+    Dict, 
+    Any, 
+    Optional,
+    Tuple, 
+    Union
+)
+import pyarrow.dataset as ds
+
+
+
+@dataclass(frozen=True)
+class RelationshipMapping:
+    foreign_key: str
+    reference_table: str
+    reference_key: str
+    label: str
+    variables: Optional[Dict[str, str]] = None
+
+
+
+@dataclass(frozen=True)
+class Nx1Mapping:
+    mapping: RelationshipMapping
+    from_entity: bool = False
+
+
+
+@dataclass(frozen=True)
+class MxNMapping:
+    mapping1: RelationshipMapping
+    mapping2: RelationshipMapping
+    label: str
+    from_first: bool = True
+
+
+
+Mapping = Union[List[Nx1Mapping], MxNMapping]
+
+
+
+@dataclass
+class CollectionMapping:
+    collection_name: str
+    mapping: Mapping = None
+    indices: Optional[List[str]] = None
+
+
+
+class DataSource(ABC):
+    def __init__(self, file_extension: str) -> None:
+        self._file_extension: str = file_extension
+
+    
+    @abstractmethod
+    def load_data(
+        self,
+        collection_name: str,
+        is_cross_table: bool=False
+    ) -> None:
+        pass
+
+
+
+class S3DataSource(DataSource):
+    def __init__(
+        self, 
+        bucket_name: str,
+        s3_access_key: str,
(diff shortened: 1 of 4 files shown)


gqlalchemy/loaders.py
```diff
@@ -0,0 +1,386 @@
...
+    foreign_key: str
+    reference_table: str
+    reference_key: str
+    label: str
+    variables: Optional[Dict[str, str]] = None
+
+
+
+@dataclass(frozen=True)
+class Nx1Mapping:
```
Josipmrden, line 50: Maybe OneToManyMapping should be a better naming convention?

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,386 @@
...
+    mapping: RelationshipMapping
+    from_entity: bool = False
+
+
+
+@dataclass(frozen=True)
+class MxNMapping:
```
Josipmrden, line 57: Maybe ManyToManyMapping should be a better naming convention?

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,386 @@
...
+    Optional,
+    Tuple, 
+    Union
+)
+import pyarrow.dataset as ds
+
+
+
+@dataclass(frozen=True)
+class RelationshipMapping:
```
Josipmrden, line 40: Maybe ForeignKeyMapping should be a better naming convention?

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,386 @@
...
+                yield batch
+
+
+@dataclass(frozen=True)
+class Configuration:
+    label: Optional[str] = None
+    column_names_mapping: Optional[Dict[str, str]] = None
+
+
+class Translator:
```
Josipmrden, line 142: This is too abstract, maybe a more descriptive name, e.g. SQLToGraphTranslator?

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,386 @@
...
+        # Load batches from raw data
+        for batch in dataset.to_batches(
+            columns=columns,
+        ):
+            for batch in batch.to_pylist():
+                yield batch
+
+
+@dataclass(frozen=True)
+class Configuration:
```
Josipmrden, line 137: Try doing docs on every parameter of the classes, otherwise we'll have a hard time from developers figuring out what it means just from class names and properties

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,386 @@
...
+        data_source: DataSource,
+        data_configuration: Dict[str, Any], 
+        memgraph: Optional[Memgraph] = None
+    ) -> None:
+        self._memgraph: Memgraph = memgraph if memgraph is not None else Memgraph()
+        self._data_source = data_source
+        self._collection_mappings: List[CollectionMapping] = []
+        self._cross_relationship_mappings: List[CollectionMapping] = []
+
+        self._load_name_mappings(data_configuration["name_mappings"])
```
Josipmrden, line 159: also rename this name_mappings

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,386 @@
...
+        self._load_nx1_relations_and_indices(data_configuration["nx1_relations"], data_configuration["indices"])
+        self._load_mxn_relations(data_configuration["mxn_relations"])
+
+
+    def translate(self, drop_database_on_start: bool):
+        if drop_database_on_start:
+            self._memgraph.drop_database()
+            self._drop_indices() # to gqla
+            self._drop_triggers() # to gqla
```
Josipmrden, line 168: Do we want to drop those triggers when finished, would it make sense to add that bool as well?

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,386 @@
...
+    def _load_nx1_relations_and_indices(self, nx1_relations: Dict[str, List[str]], indices: Dict[str, List[str]]) -> None:
+        self._collection_mappings = [
+            CollectionMapping(
+                collection_name=table_name, 
+                mapping=[Nx1Mapping(mapping=RelationshipMapping(**relation)) for relation in relations], 
+                indices=indices[table_name]
+            ) for table_name, relations in nx1_relations.items()]
+
+
+    def _load_mxn_relations(self, mxn_relations: Dict):
```
Josipmrden, line 349: Renama methods respectively with new property names

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,386 @@
...
+    from_first: bool = True
+
+
+
+Mapping = Union[List[Nx1Mapping], MxNMapping]
+
+
+
+@dataclass
+class CollectionMapping:
```
Josipmrden, line 70: The naming is kinda misleading since it doesn't have anything to do with python collections :/

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,386 @@
...
+        self._s3_session_token: Optional[str] = s3_session_token
+
+    
+    def load_data(
+        self,
+        collection_name: str,
+        is_cross_table: bool=False,
+        columns: Optional[List[str]] = None,
+    ) -> None:
+        s3 = fs.S3FileSystem(
```
Josipmrden, line 116: Can we try to do this with **kwargs? Seems like a lot of repetitive code, just to map to different properties, based on the system?

Josipmrden (changes requested):

gqlalchemy/loaders.py
BorisTasevski: This is not field name, it's method name

gqlalchemy/loaders.py
BorisTasevski: It's not when finished, it's if the flag is passed on starting of the translation, then the triggers and indices are dropped

Issue #93, opened 2022-02-14
Allow bootstrap_servers argument usage without quotes


g-despot:
When creating a stream, the `bootstrap_servers` argument needs to be specified with quotes around it:

```python
stream = MemgraphKafkaStream(
    name="ratings_stream",
    topics=["ratings"],
    transform="movielens.rating",
    bootstrap_servers="'kafka:9092'",
)
memgraph.create_stream(stream)
```

This can sometimes be misleading and users will maybe pass a plain string without the additional quotes. Both options need to be handled in the background and adjusted to the [proper Cypher syntax](https://memgraph.com/docs/memgraph/reference-guide/streams#kafka).

Additionally, passing multiple bootstrap servers should be allowed in the form of a list of strings:

```python
stream = MemgraphKafkaStream(
    name="ratings_stream",
    topics=["ratings"],
    transform="movielens.rating",
    bootstrap_servers=["kafka:9092", "kafka:9093"],
)
memgraph.create_stream(stream)
```

PR #98, opened 2022-03-07
bootstrap_servers argument usage

brunos252:
Update the `bootstrap_servers` argument to accept unquoted values and lists of unquoted values.


gqlalchemy/models.py
```diff
@@ -100,7 +100,7 @@ class MemgraphKafkaStream(MemgraphStream):
     consumer_group: str = None
     batch_interval: str = None
     batch_size: str = None
-    bootstrap_servers: str = None
+    bootstrap_servers: Union[str, List] = None
```
BorisTasevski, line 103: List of what?

gqlalchemy/models.py
```diff
@@ -113,7 +113,11 @@ def to_cypher(self) -> str:
         if self.batch_size is not None:
             query += f" BATCH_SIZE {self.batch_size}"
         if self.bootstrap_servers is not None:
-            query += f" BOOTSTRAP_SERVERS {self.bootstrap_servers}"
+            if isinstance(self.bootstrap_servers, str):
+                servers_field = f"'{self.bootstrap_servers}'"
+            else:
+                servers_field = str(self.bootstrap_servers)[1:-1]
+            query += f" BOOTSTRAP_SERVERS {servers_field}"
```
BorisTasevski, line 120: Consider changing this if-else into one-liner

BorisTasevski (commented):
Minor updates

gqlalchemy/models.py
brunos252: List of servers, as it is mentioned in the issue [https://github.com/memgraph/gqlalchemy/issues/93](url)

gqlalchemy/models.py
brunos252: can you help me understand how? Something like:
servers_field = f"'{self.bootstrap_servers}'" if isinstance(self.bootstrap_servers, str) else str(self.bootstrap_servers)[1:-1]
Seems too long

BorisTasevski (approved):

brunos252:
Build and Test fails because of pymgclient version 1.1.0 in pyproject.toml and poetry.lock, this is not an issue here as it is like this in develop branch

PR #99, opened 2022-03-08
"property" to "item" rename in "where" method and remove spaces when it is a label

brunos252:
- should there be a xor_where method?
- also, there are no tests for and_where and or_where methods
- is it possible to build a query using WHERE (n)-[:KNOWS]-({name: 'Timothy'})?

I could address some of these questions, but need confirmation (clarification?) whether to go in that direction.


gqlalchemy/query_builder.py
```diff
@@ -451,29 +451,32 @@ def from_(
 
         return self
 
-    def where(self, property: str, operator: str, value: Any) -> "DeclarativeBase":
+    def where(self, item: str, operator: str, value: Any) -> "DeclarativeBase":
         """Creates a WHERE statement Cypher partial query."""
         value_cypher = to_cypher_value(value)
+        separator = "" if operator == ":" else " "
```
BorisTasevski, line 457: Extract string constants

gqlalchemy/query_builder.py
```diff
@@ -451,29 +451,32 @@ def from_(
...
+            WhereConditionPartialQuery(WhereConditionConstants.WHERE, separator.join([item, operator, value_cypher]))
         )
 
         return self
 
-    def and_where(self, property: str, operator: str, value: Any) -> "DeclarativeBase":
+    def and_where(self, item: str, operator: str, value: Any) -> "DeclarativeBase":
         """Creates a AND (expression) statement Cypher partial query."""
         value_cypher = to_cypher_value(value)
+        separator = "" if operator == ":" else " "
```
BorisTasevski, line 467: Extract string constants

gqlalchemy/query_builder.py
```diff
@@ -451,29 +451,32 @@ def from_(
...
+            WhereConditionPartialQuery(WhereConditionConstants.AND, separator.join([item, operator, value_cypher]))
         )
 
         return self
 
-    def or_where(self, property: str, operator: str, value: Any) -> "DeclarativeBase":
+    def or_where(self, item: str, operator: str, value: Any) -> "DeclarativeBase":
         """Creates a OR (expression) statement Cypher partial query."""
         value_cypher = to_cypher_value(value)
+        separator = "" if operator == ":" else " "
```
BorisTasevski, line 477: Extract string constants; the separator is declared and then used in only one place, consider merging the two

BorisTasevski (changes requested):

gqlalchemy/query_builder.py
brunos252: Extract where?
;Something like:
        self._query.append(WhereConditionPartialQuery(
            WhereConditionConstants.WHERE, 
            ("" if operator == ":" else " ").join([item, operator, value_cypher])
            ))
?
is it not more readable as is?

katarinasupe:
Just a quick comment, before I check what you did here. All PRs should merge with develop branch, that is, you have to rebase this PR. In that way you'll be up to date with all the new changes for the new release.

katarinasupe:
It is not possible to build query is it possible to build a query using `WHERE (n)-[:KNOWS]-({name: 'Timothy'})`, since [patterns in expressions](https://memgraph.com/docs/cypher-manual/differences#patterns-in-expressions) are not supported in our implementation of Cypher.

katarinasupe:
- should there be a xor_where method?
- also, there are no tests for and_where and or_where methods
These two questions are too familiar for me, whether I asked them already, or implemented them, I'll get back to you once I check the develop branch or one of the current PRs.

gqlalchemy/query_builder.py
BorisTasevski: No, we are hard-coding this operator ":"

BorisTasevski (changes requested):
XOR WHERE and tests are not subjects of this task.

brunos252:
xor_where, and tests are already in develop branch, I didn't see it as I was based in main.
So, all my questions are answered

Josipmrden (approved):
Just a general remark, when doing PR's, the title is always in Present Tense, commits can be in past tense :D

g-despot (approved):

PR #100, opened 2022-03-09
[develop < T0035 - GA] Add table to graph importer

BorisTasevski:


gqlalchemy/loaders.py
```diff
@@ -0,0 +1,416 @@
...
+        self._s3_session_token: Optional[str] = s3_session_token
+
+    
+    def load_data(
+        self,
+        collection_name: str,
+        is_cross_table: bool = False,
+        columns: Optional[List[str]] = None,
+    ) -> None:
+        s3 = fs.S3FileSystem(
```
Josipmrden, line 116: Does it make sense to put those in some _config property of S3DataSource and then just do fs.S3FileSystem(**config), since we don't use any of those explicitly anywhere else

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,416 @@
...
+    def __init__(
+        self,
+        data_source: DataSource,
+        data_configuration: Dict[str, Any], 
+        memgraph: Optional[Memgraph] = None
+    ) -> None:
+        self._memgraph: Memgraph = memgraph if memgraph is not None else Memgraph()
+        self._data_source: DataSource = data_source
+
+        self.__load_configuration(data_configuration=data_configuration)
```
Josipmrden, line 160: Why are some methods prefixed with double underscore? One is by convention private, I'm not sure what are you trying to achieve? Is it some functionality that is done with __ when you type it in front?

Josipmrden: https://dbader.org/blog/meaning-of-underscores-in-python#:~:text=A%20double%20underscore%20prefix%20causes,the%20class%20is%20extended%20later.

BorisTasevski: I want to keep it private, so when someone inherits this class it can't invoke double-underscore-prefixed methods

gqlalchemy/loaders.py
BorisTasevski: We could, but i like type checking

BorisTasevski: However i will look into it further

Josipmrden: There is type checking if we leave the arguments as they are in _init_ and then just add to private dictionary, isn't it?

gqlalchemy/loaders.py
Josipmrden: ok sounds fine

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,416 @@
...
+@dataclass(frozen=True)
+class ManyToManyMapping:
+    mapping1: ForeignKeyMapping
+    mapping2: ForeignKeyMapping
+    label: str
+    from_first: bool = True
+
+
+
+Mapping = Union[List[OneToManyMapping], ManyToManyMapping]
```
brunos252, line 65: Mapping is actually a type in >typing< module, is this a problem?

BorisTasevski: It's not imported from typing

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,416 @@
...
+@dataclass(frozen=True)
+class OneToManyMapping:
+    mapping: ForeignKeyMapping
+    from_entity: bool = False
+
+
+
+@dataclass(frozen=True)
+class ManyToManyMapping:
+    mapping1: ForeignKeyMapping
```
Josipmrden, line 58: Can we change this to mapping_a, and mapping_b

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,416 @@
...
+    mapping2: ForeignKeyMapping
+    label: str
+    from_first: bool = True
```
Josipmrden, line 61: Do we need to specify the from_first since we essentially control which is the first and second mapping?
Then we could name it mapping_from and mapping_to?

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,416 @@
...
+    Any, 
+    Optional,
+    Tuple, 
+    Union
+)
+import pyarrow.dataset as ds
+
+
+
+@dataclass(frozen=True)
```
brunos252, line 39: PEP 8 states we should surround classes with 2 blank lines, you use 3? https://peps.python.org/pep-0008/#blank-lines

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,416 @@
...
+        self, 
+        bucket_name: str,
+        s3_access_key: str,
+        s3_secret_key: str,
+        s3_region: str,
+        file_extension: str,
+        s3_session_token: Optional[str] = None,
+    ) -> None:
+        super().__init__(file_extension=file_extension)
+        self._bucket_name: str = bucket_name
```
brunos252, line 103: do we need double type checking? in the arguments list and also here, or does this serve some other purpose?

Josipmrden:
And please resolve all conversations that you think require no further comments (i.e. you implemented changes based on suggestions, or you successfully defended your own opinion and your change stays)

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,570 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from string import Template
+
+from . import Memgraph
+from .query_builder import QueryBuilder, Unwind
```
Josipmrden, line 18: I know it's inside gqlalchemy. A preference though only.

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,570 @@
...
+    :param mapping: Foreign key used for mapping
+    :param label: Label which will be applied to the relationship created from this object
+    :param from_entity: Direction of the relationship created from mapping object
+    :param variables: Variables that will be added to the relationship created from this object (Optional)
+    """
+
+    foreign_key: ForeignKeyMapping
+    label: str
+    from_entity: bool = False
+    variables: Optional[Dict[str, str]] = None
```
g-despot, line 75: Why is it named `variables` and not `properties`?

gqlalchemy/loaders.py
katarinasupe: We should then agree on this and correct it everywhere. Besides that, we need to add that in contribution guide if we decide.

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,570 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from string import Template
+
+from . import Memgraph
```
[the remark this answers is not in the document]
brunos252: As far as i know it is recommended to use absolute imports. One case where relative may be useful is when the import statement is too long.

gqlalchemy/loaders.py
BorisTasevski: We will resolve this in subsequent PR

gqlalchemy/loaders.py
BorisTasevski: I didn't understand the comment in the first place, but You were correct. The handling is now different

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,569 @@
...
+    reference_table: str
+    reference_key: str
+
+
+@dataclass(frozen=True)
+class OneToManyMapping:
+    """
+    Class that holds the full description of a single one to many mapping in a table.
+
+    :param mapping: Foreign key used for mapping
```
brunos252, line 66: change mapping to foreign_key

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,569 @@
...
+    parameters: Optional[Dict[str, str]] = None
+
+
+@dataclass(frozen=True)
+class ManyToManyMapping:
+    """
+    Class that holds the full description of a single many to many mapping in a table.
+    Many to many mapping is intended to be used in case of associative tables
+
+    :param mapping_from: Describes the source of the relationship
```
brunos252, line 84: also here, it was renamed

gqlalchemy/loaders.py
```diff
@@ -0,0 +1,569 @@
...
+        :param collection_name: Name of the collection from which to read the data
+        :param is_cross_table: Indicate whether or not the collection contains associative table (default=False)
+        :param columns: List of columns to be read from the collection (Optional)
+        """
+        s3 = fs.S3FileSystem(**self._s3_config)
+        source = f"{self._bucket_name}/{collection_name}.{self._file_extension}"
+        print("Loading data from " + ("cross " if is_cross_table else "") + f"table {source}...")
+
+        # Load dataset via Pyarrow
+        dataset = ds.dataset(source=source, filesystem=s3)
```
brunos252, line 201: need to specify `format=file_extension` parameter, default is parquet

brunos252 (changes requested):

gqlalchemy/loaders.py
BorisTasevski: done

gqlalchemy/loaders.py
BorisTasevski: done

g-despot (approved):

Issue #88, opened 2022-01-28
Property without Field() shouldn't be saved into Memgraph


katarinasupe:
A property without `Field()` shouldn't be saved into Memgraph. 
Example: [tests/ogm/without_field_test.py](https://github.com/memgraph/gqlalchemy/pull/86/commits/b4e1d662311e03a3ade2829460dbe1f96ebfea76#diff-8e3cf8b7a44654652b1dab9ec1ff8b0d5969c5e26888547f051c754b790f69ea)
The `followers` property shouldn't be saved into Memgraph.

Mislav's suggestion -> we can make `Node` and `Relationship` class an exception because we want to be able to create a `Node` with `Node(id=7, _label="Stream")` and we want that the `id` property is saved into Memgraph.

Then loading like he suggested previously (https://github.com/memgraph/gqlalchemy/issues/79) should work, that is, the test https://github.com/memgraph/gqlalchemy/pull/86/commits/beb75194e6280ececc46039f35296852aa3c1777 should pass.

PR #103, opened 2022-03-14
Add example usage of class properties

katarinasupe:
In this PR, I wanted to check how to create private properties which won't be saved in the database and how to load a node with partial properties, that is, how to load a node with only required properties provided.

That is related to issues https://github.com/memgraph/gqlalchemy/issues/88 and https://github.com/memgraph/gqlalchemy/issues/79. I still need to figure out how to use load with any property, regardless of whether it's required.


gqlalchemy/query_builder.py
```diff
@@ -469,33 +469,42 @@ def from_(
...
                 WhereConditionConstants.WHERE, separator.join([item, operator, to_cypher_value(value)])
             )
         )
+        self._query.append(query)
 
         return self
 
     def and_where(self, item: str, operator: str, value: Any) -> "DeclarativeBase":
         """Creates a AND (expression) statement Cypher partial query."""
         separator = "" if operator == ":" else " "
```
brunos252, line 485: Then we don't need this line, maybe better to put "".join(...) and " ".join(...) in the code?

g-despot (approved):

katarinasupe:
Please ignore `where()` method in query builder here. We need to merge this before `where()` PR. This PR is just couple of new tests in the end.

gqlalchemy/query_builder.py
```diff
@@ -469,42 +469,51 @@ def from_(
     def where(self, item: str, operator: str, value: Any) -> "DeclarativeBase":
         """Creates a WHERE statement Cypher partial query."""
         separator = "" if operator == ":" else " "
-        self._query.append(
-            WhereConditionPartialQuery(
+        query = (
```
BorisTasevski, line 472: Rewrite this `if` statement. The only part that is changing is the third parameter in `join`, so move if-else there.

gqlalchemy/query_builder.py
```diff
@@ -469,42 +469,51 @@ def from_(
...
+        self._query.append(query)
 
         return self
 
     def and_where(self, item: str, operator: str, value: Any) -> "DeclarativeBase":
         """Creates a AND (expression) statement Cypher partial query."""
         separator = "" if operator == ":" else " "
-        self._query.append(
-            WhereConditionPartialQuery(
+        query = (
```
BorisTasevski, line 486: Rewrite this if statement. The only part that is changing is the third parameter in join, so move if-else there.

gqlalchemy/query_builder.py
```diff
@@ -469,42 +469,51 @@ def from_(
...
+        self._query.append(query)
 
         return self
 
     def or_where(self, item: str, operator: str, value: Any) -> "DeclarativeBase":
         """Creates a OR (expression) statement Cypher partial query."""
         separator = "" if operator == ":" else " "
-        self._query.append(
-            WhereConditionPartialQuery(
+        query = (
```
BorisTasevski, line 500: Rewrite this if statement. The only part that is changing is the third parameter in join, so move if-else there.

tests/memgraph/test_query_builder.py
```diff
@@ -320,6 +331,25 @@ def test_or_where(memgraph):
     mock.assert_called_with(expected_query)
 
 
+def test_or_where_label(memgraph):
+    query_builder = (
```
BorisTasevski, line 335: add keywords to arguments

tests/memgraph/test_query_builder.py
```diff
@@ -301,6 +301,17 @@ def test_where(memgraph):
     mock.assert_called_with(expected_query)
 
 
+def test_where_label(memgraph):
+    query_builder = QueryBuilder().match().node(variable="n").where("n", ":", "L1").return_()
```
BorisTasevski, line 305: add keywords to arguments

tests/memgraph/test_query_builder.py
```diff
@@ -339,6 +369,25 @@ def test_and_where(memgraph):
     mock.assert_called_with(expected_query)
 
 
+def test_and_where_label(memgraph):
+    query_builder = (
```
BorisTasevski, line 373: add keywords to arguments

tests/memgraph/test_query_builder.py
```diff
@@ -358,6 +407,25 @@ def test_xor_where(memgraph):
     mock.assert_called_with(expected_query)
 
 
+def test_xor_where_label(memgraph):
+    query_builder = (
```
BorisTasevski, line 411: add keywords to arguments

BorisTasevski (changes requested):
Classes as fixtures should also be discussed here

gqlalchemy/query_builder.py
katarinasupe: This is all different in new PR for WHERE clause, so please check there.

gqlalchemy/query_builder.py
katarinasupe: This is all different in new PR for WHERE clause, so please check there.

gqlalchemy/query_builder.py
katarinasupe: This is all different in new PR for WHERE clause, so please check there.

tests/memgraph/test_query_builder.py
katarinasupe: Added that to new PR for WHERE.

gqlalchemy/query_builder.py
BorisTasevski: Can we... Revert this, then?

gqlalchemy/query_builder.py
BorisTasevski: Can we... Revert this, then?

gqlalchemy/query_builder.py
BorisTasevski: Can we... Revert this, then?

gqlalchemy/query_builder.py
```diff
@@ -468,44 +468,64 @@ def from_(
...
         """Creates a WHERE statement Cypher partial query."""
-        separator = "" if operator == ":" else " "
+
+        if operator == ":":
+            separator = ""
+            cypher_value = value
+        else:
+            separator = " "
+            cypher_value = to_cypher_value(value)
+
```
Josipmrden, line 478: You repeat this 3 times, it makes sense to make some auxiliary methods then
it's fine if it's something like `_get_separator` or similar.

BorisTasevski (approved):

PR #105, opened 2022-03-16
[develop < T0020-GA] Fix label inheritance

g-despot:
- [x] Fixed behavior when labels aren't inherited from parent classes.
- [x] Added tests to cover most inheritance cases.

### DIFF
--- a/gqlalchemy/models.py
+++ b/gqlalchemy/models.py
@@ -350,19 +350,22 @@ class NodeMetaclass(BaseModel.__class__):
 
             return None
 
+        def get_base_labels() -> Set[str]:
+            base_labels = set()
+            nonlocal bases
+            for base in bases:
+                base_labels = base_labels.union(base.labels if hasattr(base, "labels") else set())
+            return base_labels
+
         cls = super().__new__(mcs, name, bases, namespace, **kwargs)
+
         cls.label = kwargs.get("label", name)
         if name == "Node":
             pass
-        elif "labels" in kwargs:  # overrides superclass labels
-            labs = set()
-            for base in bases:
-                labs = labs.union(base.labels if hasattr(base, "labels") else set())
-            cls.labels = {cls.label} | kwargs["labels"] | labs
-        elif hasattr(cls, "labels"):
-            cls.labels = cls.labels | {cls.label}
+        elif "labels" in kwargs:
+            cls.labels = {cls.label} | kwargs["labels"] | get_base_labels()
         else:
-            cls.labels = {cls.label}
+            cls.labels = {cls.label} | get_base_labels()
 
         for field in cls.__fields__:
             attrs = cls.__fields__[field].field_info.extra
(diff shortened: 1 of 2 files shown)


katarinasupe (approved):
Great job on testing!

katarinasupe:
I just reformatted one line with black, and tests should pass now.

katarinasupe (approved):

gqlalchemy/models.py
```diff
@@ -350,16 +350,22 @@ def field_in_superclass(field, constraint):
 
             return None
 
+        def get_base_labels() -> Set[str]:
+            base_labels = set()
+            nonlocal bases
+            for base in bases:
+                base_labels = base_labels.union(base.labels if hasattr(base, "labels") else set())
```
BorisTasevski, line 357: feels like this union operation could be skipped if `hasattr` returns `False`

BorisTasevski (commented):
Maybe a minor improvement.

gqlalchemy/models.py
g-despot: Thank you for the suggestion. I skipped the union operation if `hasattr` returns `False`.

gqlalchemy/models.py
```diff
@@ -354,7 +354,7 @@ def get_base_labels() -> Set[str]:
             base_labels = set()
             nonlocal bases
             for base in bases:
-                base_labels = base_labels.union(base.labels if hasattr(base, "labels") else set())
+                base_labels = base_labels.union(base.labels) if hasattr(base, "labels") else base_labels
```
BorisTasevski, line 357: `if hasattr(base, "labels"): base_labels.union(base.labels)` ?

g-despot: Fixed

gqlalchemy/models.py
```diff
@@ -350,16 +350,23 @@ def field_in_superclass(field, constraint):
...
+            nonlocal bases
+            for base in bases:
+                if hasattr(base, "labels"):
+                    base_labels = base_labels.union(base.labels)
+            return base_labels
+
         cls = super().__new__(mcs, name, bases, namespace, **kwargs)
+
         cls.label = kwargs.get("label", name)
         if name == "Node":
```
BorisTasevski, line 364:
```
if name != "Node":
     cls.labels = {cls.label} | get_base_labels() | kwargs.get("labels", set())
```

?

BorisTasevski (changes requested):
There are conflicts as well

gqlalchemy/models.py
BorisTasevski: And another Question, `|` operator is a shorthand for `union()` method, so why are we using both?

g-despot: Thank you for the suggestion, I replaced the block with one if statement and replaced the `|` operator with the `union()` method.

gqlalchemy/models.py
```diff
@@ -356,17 +356,13 @@ def get_base_labels() -> Set[str]:
...
-
         cls.label = kwargs.get("label", name)
-        if name == "Node":
-            pass
-        elif "labels" in kwargs:
-            cls.labels = {cls.label} | kwargs["labels"] | get_base_labels()
-        else:
-            cls.labels = {cls.label} | get_base_labels()
+        if name != "Node":
+            cls.labels = get_base_labels().union({cls.label}, kwargs.get("labels", set()))
```
katarinasupe, line 371: I like the way you changed this 😄

BorisTasevski: Thanks :)

BorisTasevski (approved):

PR #111, opened 2022-03-28
[develop < T0039-GA] Change where procedure in query builder

katarinasupe:
### Description

Changed `where()`, `and_where()`, `or_where()` and `xor_where()` methods (added keyword arguments). Now label can be used as a left-hand side value in where clause.
Added tests for all methods.

### Pull request type

- [x] Bugfix 
- [x] Refactoring with functional or API changes
- [x] Documentation content changes

### Related issues

Closes https://github.com/memgraph/gqlalchemy/issues/74

######################################

### Reviewer checklist (the reviewer checks this part)
- [ ] Core feature implementation
- [ ] Tests
- [ ] Code documentation
- [ ] Documentation on [memgraph/docs](https://github.com/memgraph/docs)

######################################

PR #114, opened 2022-03-30
[develop < T0046-GA] Add order by descending

katarinasupe:
### Description

What's new:
- Exceptions for ordering.
- Enums for Where (WHERE, AND, OR, XOR, NOT) and Order (ASC, ASCENDING, DESC, DESCENDING).
- All where methods (where, and, or, xor) use their classes to construct queries. Those classes inherit from the `WhereConditionPartialQuery` class. 
- All where_not methods (where_not, and_not, or_not, xor_not) use their classes to construct queries. Those classes inherit from the `WhereNotConditionPartialQuery` class, which inherits from the `WhereConditionPartialQuery` class.
- There is one order_by method that allows either property (default ascending order), a tuple of property and order, or a list of either property or tuple of property and order.

### Pull request type

- [x] Feature

- [ ] Tests
- [ ] Code documentation

### DIFF
--- a/gqlalchemy/query_builder.py
+++ b/gqlalchemy/query_builder.py
@@ -146,7 +146,7 @@ class CallPartialQuery(PartialQuery):
 
 
 class WhereConditionPartialQuery(PartialQuery):
-    def __init__(self, item: str, operator: str, keyword: str = "", **kwargs):
+    def __init__(self, item: str, operator: str, keyword: str = None, **kwargs):
         super().__init__(type=DeclarativeBaseTypes.WHERE)
 
         self.keyword = keyword
@@ -154,8 +154,8 @@ class WhereConditionPartialQuery(PartialQuery):
 
     def construct_query(self) -> str:
         """Constructs a where partial query."""
-        item = self.keyword if self.keyword != "" else self.type
-        return f" {item} {self.query} "
+
+        return f" {self.keyword if self.keyword is not None else self.type} {self.query} "
 
     def _build_where_query(self, item: str, operator: str, **kwargs) -> "DeclarativeBase":
         """Builds parts of a WHERE Cypher query divided by the boolean operators."""
--- a/gqlalchemy/query_builder.py
+++ b/gqlalchemy/query_builder.py
@@ -14,11 +14,12 @@
 
 import re
 from abc import ABC, abstractmethod
-from typing import Any, Dict, Iterator, List, Optional, Union
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
 
 from .memgraph import Connection, Memgraph
 from .utilities import to_cypher_labels, to_cypher_properties, to_cypher_value
 from .models import Node, Relationship
+from .exceptions import GQLAlchemyMissingOrdering
 
 
 class DeclarativeBaseTypes:
@@ -33,7 +34,6 @@ class DeclarativeBaseTypes:
     MERGE = "MERGE"
     NODE = "NODE"
     ORDER_BY = "ORDER_BY"
-    ORDER_BY_DESC = "ORDER_BY_DESC"
     OR_WHERE = "OR_WHERE"
     REMOVE = "REMOVE"
     RETURN = "RETURN"
@@ -61,6 +61,17 @@ class WhereConditionConstants:
     AND = "AND"
     OR = "OR"
     XOR = "XOR"
+
+
+class OrderByConstants:
+    ASC = "ASC"
+    ASCENDING = "ASCENDING"
+    DESC = "DESC"
+    DESCENDING = "DESCENDING"
+
+    @classmethod
+    def is_member(cls, ordering: str) -> bool:
+        return ordering in [cls.ASC, cls.ASCENDING, cls.DESC, cls.DESCENDING]
 
 
 class NoVariablesMatchedException(Exception):
@@ -327,23 +338,48 @@ class ReturnPartialQuery(PartialQuery):
 
 
 class OrderByPartialQuery(PartialQuery):
-    def __init__(self, properties: str):
+    def __init__(self, properties: Union[str, Tuple[str, str], List[Union[str, Tuple[str, str]]]]):
         super().__init__(DeclarativeBaseTypes.ORDER_BY)
 
         self.properties = properties
 
     def construct_query(self) -> str:
         """Creates a ORDER BY statement Cypher partial query."""
-        return f" ORDER BY {self.properties} "
 
+        result_query = " ORDER BY "
+        if isinstance(self.properties, str):
+            result_query += f"{self._read_str(self.properties)} "
+        elif isinstance(self.properties, tuple):
+            result_query += f"{self._read_tuple(self.properties)} "
+        elif isinstance(self.properties, list):
+            result_query += f"{self._read_list(self.properties)} "
+        else:
+            raise GQLAlchemyMissingOrdering
 
-class OrderByDescPartialQuery(OrderByPartialQuery):
-    def __init__(self, properties: str):
-        super().__init__(properties)
+        return result_query
 
-    def construct_query(self) -> str:
-        """Creates an ORDER BY DESC statement Cypher partial query."""
-        return super().construct_query() + "DESC "
+    def _read_str(self, property: str) -> str:
+        return f"{property}"
+
+    def _read_tuple(self, property: Tuple[str, str]) -> str:
+        if not OrderByConstants.is_member(property[1]):
+            raise GQLAlchemyMissingOrdering
+        return f"{property[0]} {property[1]}"
+
+    def _read_list(self, property: List[Union[str, tuple[str, str]]]):
+        result = ""
+        for i, item in enumerate(property):
+            if isinstance(item, str):
+                result += self._read_str(item)
+            elif isinstance(item, tuple):
+                result += self._read_tuple(item)
+            else:
+                raise GQLAlchemyMissingOrdering
+
+            if i != len(property) - 1:
+                result += ", "
+
+        return result
 
 
 class LimitPartialQuery(PartialQuery):
@@ -583,15 +619,9 @@ class DeclarativeBase(ABC):
 
         return self
 
-    def order_by(self, properties: str) -> "DeclarativeBase":
+    def order_by(self, properties: Union[str, Tuple[str, str], List[Union[str, Tuple[str, str]]]]) -> "DeclarativeBase":
         """Creates a ORDER BY statement Cypher partial query."""
         self._query.append(OrderByPartialQuery(properties))
-
-        return self
-
-    def order_by_desc(self, properties: str) -> "DeclarativeBase":
-        """Creates an ORDER BY DESC statement Cypher partial query."""
-        self._query.append(OrderByDescPartialQuery(properties))
 
         return self
 
--- a/gqlalchemy/query_builder.py
+++ b/gqlalchemy/query_builder.py
@@ -550,6 +550,22 @@ class DeclarativeBase(ABC):
 
         Returns:
             self: A partial Cypher query built from the given parameters.
+
+        Examples:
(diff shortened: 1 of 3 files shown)


gqlalchemy/query_builder.py
```diff
@@ -26,6 +26,7 @@ class DeclarativeBaseTypes:
     CALL = "CALL"
     CREATE = "CREATE"
     DELETE = "DELETE"
+    DESC = "DESC"
```
g-despot, line 29: Why did you add it here?

tests/memgraph/test_query_builder.py
```diff
@@ -428,6 +428,16 @@ def test_orderby_desc(memgraph):
     mock.assert_called_with(expected_query)
 
 
+def test_order_by_desc(memgraph):
+    query_builder = QueryBuilder().match().node(variable="n").return_().order_by_desc("n.id")
```
BorisTasevski, line 432: key for parameter in `order_by_desc` omitted

BorisTasevski (changes requested):

gqlalchemy/query_builder.py
```diff
@@ -336,6 +337,17 @@ def construct_query(self) -> str:
...
 
+class OrderByDescPartialQuery(PartialQuery):
+    def __init__(self, properties: str):
+        super().__init__(DeclarativeBaseTypes.DESC)
+
+        self.properties = properties
+
+    def construct_query(self) -> str:
+        """Creates a ORDER BY statement Cypher partial query."""
+        return OrderByPartialQuery(self.properties).construct_query() + "DESC "
```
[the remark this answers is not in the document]
katarinasupe: Because it needs to have one space extra at the end, I will try solving that in another way. I'll let you know once I fix it.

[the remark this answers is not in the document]
BorisTasevski: I feel that most of the code needs to be rewritten.

I would write something like:

`return f"{OrderByPartialQuery(self.properties).construct_query()}{DeclarativeBaseTypes.DESC} "`

gqlalchemy/query_builder.py
katarinasupe: Okay, my fault, totally did this wrong

gqlalchemy/query_builder.py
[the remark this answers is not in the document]
katarinasupe: As Ivan noticed above, I was wrong with putting the DESC in DeclarativeBaseTypes. Therefore, we don't have constants for any of the strings in the query. That's why I need to use "DESC ", like "ORDER BY "which was used above in `order_by()` method. But I agree, this will need to be changed sometimes.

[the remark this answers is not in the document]
katarinasupe: With the current code base, the correct would be:

`return f"{OrderByPartialQuery(self.properties).construct_query()}DESC "`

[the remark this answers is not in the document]
katarinasupe: I pushed new commit, let me know if that's fine.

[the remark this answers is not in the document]
katarinasupe: I pushed the version with the inheritance, we can discuss it tomorrow. If it's not okay, I can revert it to the original version.

tests/memgraph/test_query_builder.py
```diff
@@ -428,6 +429,63 @@ def test_orderby_desc(memgraph):
...
+
+    with patch.object(Memgraph, "execute_and_fetch", return_value=None) as mock:
+        query_builder.execute()
+
+    mock.assert_called_with(expected_query)
+
+
+def test_order_by_wrong_ordering(memgraph):
+    with pytest.raises(GQLAlchemyMissingOrdering):
+        QueryBuilder().match().node(variable="n").return_().order_by(("n.id", "DESCE")).execute()
```
katarinasupe, line 444: It is not clear to me why `GQLAlchemyMissingOrdering` error is being raised only if I add `execute() `at the end of the query builder. I did [a similar thing in `where()`](https://github.com/memgraph/gqlalchemy/pull/111/files#diff-b58f85388321491b6930511b2eae02b960636ea7c5b2d2ab20b465dc735471f6R344) clause and `execute()` wasn't necessary.

BorisTasevski: Because we are doing checks at the time of the execution and not in creation time.

gqlalchemy/query_builder.py
```diff
@@ -609,7 +640,9 @@ def return_(self, results: Optional[Dict[str, str]] = {}) -> "DeclarativeBase":
 
         return self
 
-    def order_by(self, properties: str) -> "DeclarativeBase":
+    def order_by(
+        self, properties: Union[str, Tuple[str, Order], List[Union[str, Tuple[str, Order]]]]
+    ) -> "DeclarativeBase":
```
brunos252, line 645: language question, why do we need qoutes around DeclarativeBase?

gqlalchemy/query_builder.py
```diff
@@ -330,14 +340,35 @@ def construct_query(self) -> str:
...
+    def __init__(self, properties: Union[str, Tuple[str, Order], List[Union[str, Tuple[str, Order]]]]):
         super().__init__(DeclarativeBaseTypes.ORDER_BY)
 
         self.properties = properties
 
     def construct_query(self) -> str:
         """Creates a ORDER BY statement Cypher partial query."""
-        return f" ORDER BY {self.properties} "
+
+        return f" {self.type} {(self._read_list(self.properties)) if isinstance(self.properties, list) else self._read_item(self.properties)} "
```
brunos252, line 351: This line is really long, [PEP8](https://peps.python.org/pep-0008/#maximum-line-length) says it should be up to 80 characters, but in our checks flake8 permits it, Why? Also, should we avoid such long lines for readability?

brunos252:
Do you plan on making more comments in the code?

gqlalchemy/query_builder.py
BorisTasevski: https://peps.python.org/pep-0484/#forward-references

gqlalchemy/query_builder.py
BorisTasevski: Fixed

katarinasupe:
@brunos252  I think the plan is to document everything better here: https://github.com/memgraph/gqlalchemy/pull/116

gqlalchemy/query_builder.py
```diff
@@ -139,15 +146,46 @@ def construct_query(self) -> str:
...
 
 class WhereConditionPartialQuery(PartialQuery):
-    def __init__(self, keyword: str, query: str):
-        super().__init__(DeclarativeBaseTypes.WHERE)
+    def __init__(self, item: str, operator: str, keyword: str = None, **kwargs):
+        super().__init__(type=DeclarativeBaseTypes.WHERE)
 
         self.keyword = keyword
-        self.query = query
+        self._build_where_query(item=item, operator=operator, **kwargs)
```
katarinasupe, line 153: If this method is called in init and not in construct query, the type checks are done before execution?

gqlalchemy/query_builder.py
```diff
@@ -139,15 +146,46 @@ def construct_query(self) -> str:
...
+                raise GQLAlchemyLiteralAndExpressionMissingInWhere
+
+            value = to_cypher_value(literal)
+        elif literal is not None:
+            raise GQLAlchemyExtraKeywordArgumentsInWhere
+
+        self.query = ("" if operator == ":" else " ").join([item, operator, value])
+
+
+class AndWhereConditionPartialQuery(WhereConditionPartialQuery):
```
katarinasupe, line 176: This is much better 😄

gqlalchemy/query_builder.py
```diff
@@ -374,7 +437,7 @@ def construct_query(self) -> str:
 
 class DeclarativeBase(ABC):
     def __init__(self, connection: Optional[Union[Connection, Memgraph]] = None):
-        self._query: List[Any] = []
+        self._query: List[PartialQuery] = []
```
katarinasupe, line 440: Nice, we obviously missed this.

gqlalchemy/query_builder.py
```diff
@@ -609,9 +658,23 @@ def return_(self, results: Optional[Dict[str, str]] = {}) -> "DeclarativeBase":
...
+        Args:
+            properties: Properties and order by which the query results will be ordered.
+
+        Raises:
+            GQLAlchemyOrderByTypeError: Raises an error when the given ordering is of the wrong type.
+            GQLAlchemyMissingOrdering: Raises an error when the given property is neither string nor tuple.
+
+        Returns:
+            self: A partial Cypher query built from the given parameters.
+        """
```
g-despot, line 675: I would add this additional comment:
```
Examples:
            Ordering query results by the property `n.name` in ascneding order
            and by the property `n.last_name` in descending order:

            Python: `match().node(variable="n").return_().order_by(properties=["n.name", ("n.last_name", Order.DESC)])`
            Cypher: `MATCH (n) RETURN * ORDER BY n.id DESC, n.name, n.last_name DESC;`
```

g-despot (approved):

gqlalchemy/query_builder.py
BorisTasevski: Yes, the plan is to validate the query at a construction time

brunos252 (approved):

gqlalchemy/query_builder.py
```diff
@@ -139,15 +147,68 @@ def construct_query(self) -> str:
...
-        self.keyword = keyword
-        self.query = query
+    def __init__(self, item: str, operator: str, keyword: Where = Where.WHERE, is_negated: bool = False, **kwargs):
+        super().__init__(type=keyword.name if not is_negated else f"{keyword.name} {Where.NOT.name}")
+        self.query = self._build_where_query(item=item, operator=operator, **kwargs)
 
     def construct_query(self) -> str:
         """Constructs a where partial query."""
-        return f" {self.keyword} {self.query} "
+
```
Josipmrden, line 160: Delete this line. Or we discuss whether we keep the empty line after docstring.

gqlalchemy/query_builder.py
```diff
@@ -59,11 +59,19 @@ class MatchConstants:
...
+    OR = 3
+    XOR = 4
+    NOT = 5
+
+
+class Order(Enum):
+    ASC = 1
+    ASCENDING = 2
+    DESC = 3
+    DESCENDING = 4
```
Josipmrden, line 74: Do we need 4 of them? Can't we use 2?

gqlalchemy/query_builder.py
```diff
@@ -470,30 +556,50 @@ def from_(
...
+            Filtering query results by the node label.
+
+            Python: `match().node(variable="n").where(item="n", operator=":", expression="User").return_()`
+            Cypher: `MATCH (n) WHERE n:User RETURN *;`
+
+            Filtering query results by the comparison of node property and literal.
+
+            Python: `match().node(variable="n").where(item="n.age", operator=">", literal=18).return_()`
+            Cypher: `MATCH (n) WHERE n.age > 18 RETURN *;`
+        """
```
Josipmrden, line 592: Nice docs :+1:

BorisTasevski (approved):

Josipmrden (commented):
Few comments left

gqlalchemy/query_builder.py
BorisTasevski: All of them are valid

gqlalchemy/query_builder.py
katarinasupe: Thanks 😄  @g-despot made me do it 😂

gqlalchemy/query_builder.py
katarinasupe: There are a couple of occurrences of this, so let's discuss it. I'll put this in the channel.

PR #128, opened 2022-04-08
[develop < T0059-GA] Add SET method to the query builder

### LINKED ISSUE #126: [ENHANCEMENT] No support for SET Keyword
Query Builder has no support for SET keyword, other than `add_custom_cypher` method

BorisTasevski:
### Description

Added SET method to query builder.

### Related Issues

Closes #126 

### Pull request type

- [X] Feature
- [X] Code style update (formatting, renaming)
- [X] Refactoring with functional or API changes

- [ ] Tests
- [ ] Code documentation

### DIFF
--- a/gqlalchemy/exceptions.py
+++ b/gqlalchemy/exceptions.py
@@ -50,7 +50,7 @@ TypeError: The argument provided is of wrong type. Please provide str, tuple[str
 """
 
 LITERAL_AND_EXPRESSION_MISSING = """
-Can't create {clause} query without providing either 'literal' or 'expression' keyword arguments, 
+Can't create {clause} query without providing either 'literal' or 'expression' keyword arguments,
 that can be literals, labels or properties.
 """
 
--- a/gqlalchemy/exceptions.py
+++ b/gqlalchemy/exceptions.py
@@ -53,15 +53,16 @@ TypeError: The argument provided is of wrong type. Please provide str, tuple[str
 
 LITERAL_AND_EXPRESSION_MISSING = Template(
     """
-    Can't create $clause query without providing either 'literal' or 'expression' keyword arguments, that can be literals, labels or properties.
-    """
+Can't create $clause query without providing either 'literal' or 'expression' keyword arguments, 
+that can be literals, labels or properties.
+"""
 )
 
 EXTRA_KEYWORD_ARGUMENTS = Template(
     """
-    Can't create $clause query with extra keyword arguments:
-    Please provide a value to either 'literal' or 'expression' keyword arguments.
-    """
+Can't create $clause query with extra keyword arguments:
+Please provide a value to either 'literal' or 'expression' keyword arguments.
+"""
 )
 
 
--- a/gqlalchemy/exceptions.py
+++ b/gqlalchemy/exceptions.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from enum import Enum
 
 DATABASE_MISSING_IN_FIELD_ERROR_MESSAGE = """
 Can't have an index on a property without providing the database `db` object.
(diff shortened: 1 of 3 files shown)


gqlalchemy/query_builder.py
```diff
@@ -74,6 +77,12 @@ class Order(Enum):
     DESCENDING = 4
 
 
+class Operator(Enum):
+    ASSIGNMENT = "="
+    INCREMENT = "+="
+    LABEL_FILTER = ":"
+
+
```
katarinasupe, line 85: It's good that you have created Operator enum, but should we then add all possible operators for comparison here (e.g. >, <,..)? They could be used then by the user when passing an operator to the where clause.

katarinasupe: Also, should the assignment and label_filter operator then be added in where tests? Also, in where method in query builder _LABEL_FILTER = ":" is used inside the class, rather than from Operator enum. We should maybe fix this accordingly.

gqlalchemy/query_builder.py
```diff
@@ -870,6 +908,11 @@ def get_single(self, retrieve: str) -> Any:
             return result[retrieve]
         return result
 
+    def set_(self, item: str, operator: Operator, **kwargs):
+        self._query.append(SetPartialQuery(item=item, operator=operator, **kwargs))
+
+        return self
+
```
katarinasupe, line 915: Missing docstring

tests/memgraph/test_query_builder.py
```diff
@@ -1262,6 +1264,52 @@ def test_add_string_complete(memgraph):
     mock.assert_called_with(expected_query)
 
 
+def test_set_label(memgraph):
```
katarinasupe, line 1267: Check out what you can do with SET clause and then maybe add a couple of more tests: https://memgraph.com/docs/cypher-manual/clauses/set

gqlalchemy/query_builder.py
katarinasupe: I will answer myself 😹 . This `Operator` is not the same thing as comparison expressions for `WHERE` clause. Therefore, this is enough for `SET` clause.

g-despot: I would call it `SetOperator` as it's more descriptive.

gqlalchemy/exceptions.py
```diff
@@ -109,13 +111,33 @@ def __init__(self):
...
+    def __init__(self, clause: str):
         super().__init__()
-        self.message = LITERAL_AND_EXPRESSION_MISSING_IN_WHERE
+        self.message = LITERAL_AND_EXPRESSION_MISSING.format(clause=clause)
 
 
-class GQLAlchemyExtraKeywordArgumentsInWhere(GQLAlchemyError):
+class GQLAlchemyLiteralAndExpressionMissingInWhere(GQLAlchemyLiteralAndExpressionMissingInClause):
     def __init__(self):
+        super().__init__(clause="WHERE")
```
Josipmrden, line 122: does it make sense exposing these keywords to a separate enum? and then importing it. for example

class QueryConstants(Enum):
WHERE = "WHERE"

Josipmrden: and so on...

Josipmrden (approved):

Josipmrden (commented):
Minor comment

gqlalchemy/exceptions.py
katarinasupe: Oh, you meant in exceptions :D Sure, it makes sense to me. We have something similar in the query_builder, but to avoid circular imports, I'll create a new enum in exceptions and you can let me know what you think.

katarinasupe: @Josipmrden added, please review

brunos252 (approved):

PR #139, opened 2022-04-21
[develop < T0070-GA] Add more clauses for base classes

### LINKED ISSUE #138: Add more clauses for starting queries
Currently, to start building a query, aside from QueryBuilder() we can also start with:
call = Call
create = Create
match = Match
merge = Merge
unwind = Unwind
with_ = With

for instance, create().node()...

Issue is: we should add more of them, load_csv, foreach...

brunos252:
### Description

Added load_csv and return to base classes list.
Discuss:
REMOVE, DELETE and SET  are also supported per cypher grammar to be at the start of a query, but I have not included them as in agreement with Ivan they do not make much sense to be used like that, without some variable defined with MATCH in front of them.

Note:
FOREACH will also be in the base classes, but I decided to add it to the FOREACH clause PR, so this PR doesn't have to wait for that one.

### Pull request type

- [x] Bugfix
- [x] Feature

### Related issues

Closes #138 

### Checklist:

- [ ] Tests
- [ ] Code documentation

### DIFF
--- a/gqlalchemy/query_builder.py
+++ b/gqlalchemy/query_builder.py
@@ -1129,7 +1129,9 @@ class LoadCsv(DeclarativeBase):
 
 
 class Return(DeclarativeBase):
-    def __init__(self, results: Optional[Dict[str, str]] = {}, connection: Optional[Union[Connection, Memgraph]] = None):
+    def __init__(
+        self, results: Optional[Dict[str, str]] = {}, connection: Optional[Union[Connection, Memgraph]] = None
+    ):
         super().__init__(connection)
         self._query.append(ReturnPartialQuery(results))
         self._fetch_results = True
--- a/gqlalchemy/query_builder.py
+++ b/gqlalchemy/query_builder.py
@@ -1120,3 +1120,9 @@ class With(DeclarativeBase):
     ):
         super().__init__(connection)
         self._query.append(WithPartialQuery(results))
+
+
+class LoadCSV(DeclarativeBase):
+    def __init__(self, path: str, header: bool, row: str, connection: Optional[Union[Connection, Memgraph]] = None):
+        super().__init__(connection)
+        self._query.append(LoadCsvPartialQuery(path, header, row))
--- a/gqlalchemy/query_builder.py
+++ b/gqlalchemy/query_builder.py
@@ -1122,7 +1122,14 @@ class With(DeclarativeBase):
         self._query.append(WithPartialQuery(results))
 
 
-class LoadCSV(DeclarativeBase):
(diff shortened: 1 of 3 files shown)


gqlalchemy/query_builder.py
```diff
@@ -1120,3 +1120,9 @@ def __init__(
     ):
         super().__init__(connection)
         self._query.append(WithPartialQuery(results))
+
+
+class LoadCSV(DeclarativeBase):
+    def __init__(self, path: str, header: bool, row: str, connection: Optional[Union[Connection, Memgraph]] = None):
+        super().__init__(connection)
+        self._query.append(LoadCsvPartialQuery(path, header, row))
```
katarinasupe, line 1128: We have LoadCSV above and then LoadCsvPartialQuery. We should be more consistent and decide on one of those two. I vote for LoadCSVPartialQuery 😄

katarinasupe (changes requested):

katarinasupe:
And discussion-wise, I think we should follow Cypher grammar as much as we can, especially because of the next version of the query builder. I vote for implementing all clauses that can be first in the query to be base classes.

tests/memgraph/test_query_builder.py
```diff
@@ -1232,6 +1245,26 @@ def test_base_class_with(memgraph):
...
+    expected_query = " LOAD CSV FROM 'path/to/my/file.csv' WITH HEADER AS row RETURN * "
+
+    with patch.object(Memgraph, "execute_and_fetch", return_value=None) as mock:
+        query_builder.execute()
+
+    mock.assert_called_with(expected_query)
+
+
+def test_base_class_return(memgraph):
+    query_builder = return_({"n": "n"})
```
katarinasupe, line 1259: Now I started thinking about this and I returned to the `dict_to_alias_statement` method in the query builder. As I saw there, this is the only way to create `RETURN n` statement and I was surprised. Do you think it would be better if `results` in the `RETURN` clause could be either a simple `string` or `Dict` of strings? Then if it's a simple string, we just do `RETURN n`, and if it's not, the check stays the same as it is now.

brunos252: Here is the issue for this: https://github.com/memgraph/gqlalchemy/issues/140

Josipmrden (approved):

katarinasupe:
I already reviewed this and you made the necessary changes, so I think that's it! Besides failing tests, this is fine 😄

katarinasupe (approved):

katarinasupe:
I see `status:discuss`. Is there anything else that we need to discuss here? @brunos252

PR #130, opened 2022-04-12
[develop < T0026-GA] Wrap query modules

brunos252:
### Description

Wrapper class for query modules.

### Pull request type

- [x] Feature

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

######################################

### Reviewer checklist (the reviewer checks this part)
- [ ] Core feature implementation
- [ ] Tests
- [ ] Code documentation
- [ ] Documentation on [memgraph/docs](https://github.com/memgraph/docs)

######################################

Issue #123, opened 2022-04-04
Rename edge_label to edge_type


katarinasupe:
I noticed that `to()` method in query builder has argument `edge_label` for relationship type. 
```
 def to(
        self,
        edge_label: Optional[str] = "",
        directed: Optional[bool] = True,
        variable: Optional[str] = None,
        relationship: Optional["Relationship"] = None,
        **kwargs,
    ) -> "DeclarativeBase":
```

I know there was a discussion because whether to name it `type`. Type is not a reserved keyword in Python, but it is still not wise to use it. Hence, I would suggest `edge_type`.

**Edit**:
When I think about it more, it should be named `relationship` or `relationship_type`, since we are not using `edge` in our codebase.

brunos252:
relationship_type vote!

g-despot:
I also think that `relationship_type` is the most descriptive and correct naming.

brunos252:
While to() uses a variable `type_str` in:
`type_str = to_cypher_labels(edge_label)`
from_() uses `labels_str` for the same code, so this should be fixed along with other changes.

Also, `EdgePartialQuery` has a parameter called `labels` (Probably because `NodePartialQuery` has the same parameters).

PR #141, opened 2022-05-04
[develop < T0071-GA-bfs-and-shortest-path] added base class and integrated into edge query

brunos252:
### Description

Please briefly explain the changes you made here.

### Pull request type

- [x] Feature
- [x] Code style update (formatting, renaming)

### Checklist:

- [ ] Tests
- [ ] Code documentation

### DIFF
--- a/gqlalchemy/memgraph.py
+++ b/gqlalchemy/memgraph.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from abc import ABC
 import os
 import sqlite3
 from typing import Any, Dict, Iterator, List, Optional, Union
@@ -543,3 +544,44 @@ class Memgraph:
         )
 
         return self.get_variable_assume_one(results, "relationship")
+
+
+class IntegratedAlgorithm(ABC):
+    """Abstract class modeling in-Memgraph graph algorithms.
+
+    These algorithms are integrated into Memgraph codebase and are called
+    within an edge part of a query. For instance:
+    MATCH p = (:City {name: "Paris"})
+          -[:Road * bfs (e, v | e.length <= 200 AND v.name != "Metz")]->
+          (:City {name: "Berlin"})
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    def __str__(self) -> str:
+        """Instance of IntegratedAlgorithm extended object is used as a string.
+
+        Raises:
+            NotImplementedError: Inheriting class did not define its string
+            representation
+        """
+        raise NotImplementedError("Algorithm should define its str representation")
+
+    @staticmethod
+    def get_lambda(expression: str) -> str:
+        """Method for creating a general lambda expression.
+
+        Variables e and v stand for edge and vertex. The expression is used
+        e.g. for a filter lambda, to use only edges of length less than 200:
+            expression="e.length < 200"
+        with the filter lambda being:
+            (e, v | e.length < 200)
+
+        Args:
+            expression: lambda conditions or statements
+        """
+        if expression is None:
+            return ""
+        else:
+            return f"(e, v | {expression})"
--- a/gqlalchemy/memgraph.py
+++ b/gqlalchemy/memgraph.py
@@ -560,6 +560,7 @@ class IntegratedAlgorithm(ABC):
     @abstractmethod
     def __str__(self) -> str:
         """Instance of IntegratedAlgorithm subclass is used as a string"""
+        pass
 
     @staticmethod
     def to_cypher_lambda(expression: str) -> str:
--- a/gqlalchemy/memgraph.py
+++ b/gqlalchemy/memgraph.py
@@ -12,13 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from abc import ABC
 import os
 import sqlite3
+
+from abc import ABC
 from typing import Any, Dict, Iterator, List, Optional, Union
 
 from .connection import Connection
 from .disk_storage import OnDiskPropertyDatabase
+from .exceptions import (
+    GQLAlchemyError,
+    GQLAlchemyUniquenessConstraintError,
+    GQLAlchemyOnDiskPropertyDatabaseNotDefinedError,
+)
 from .models import (
     MemgraphConstraint,
     MemgraphConstraintExists,
@@ -30,11 +36,6 @@ from .models import (
     Relationship,
 )
 
-from .exceptions import (
-    GQLAlchemyError,
-    GQLAlchemyUniquenessConstraintError,
-    GQLAlchemyOnDiskPropertyDatabaseNotDefinedError,
-)
 
 __all__ = ("Memgraph",)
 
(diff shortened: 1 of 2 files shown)


Josipmrden (approved):

gqlalchemy/query_builder.py
```diff
@@ -573,13 +586,27 @@ def to(
             raise InvalidMatchChainException()
 
         if relationship is None:
-            type_str = to_cypher_labels(edge_label)
+            labels_str = to_cypher_labels(edge_label)
```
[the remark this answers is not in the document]
katarinasupe: Okay, I guess this is related to https://github.com/memgraph/gqlalchemy/issues/123
I would leave `type_str` because of this, edge does not have a label, but type and I think we should change that.

gqlalchemy/memgraph.py
```diff
@@ -543,3 +544,44 @@ def create_relationship(self, relationship: Relationship) -> Optional[Relationsh
...
+        """Instance of IntegratedAlgorithm extended object is used as a string.
+
+        Raises:
+            NotImplementedError: Inheriting class did not define its string
+            representation
+        """
+        raise NotImplementedError("Algorithm should define its str representation")
+
+    @staticmethod
+    def get_lambda(expression: str) -> str:
```
katarinasupe, line 572: Would it be better to call it `to_cypher_lambda()`, since we already have `to_cypher_value()`, `to_cypher_properties()`, etc.?

gqlalchemy/query_builder.py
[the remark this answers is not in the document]
brunos252: Great comment! I put this before the discussions, because the `to()` argument is named `edge_label`, and more importantly, `EdgePartialQuery` accepts `labels``(=labels_str). And also, `from_()` method right below `to()` uses labels_str, so a quick fix seemed appropriate. 
I will revert this and leave my comments on the issue, so we can adjust that separatly.

gqlalchemy/memgraph.py
brunos252: Sure!

katarinasupe (approved):

brunos252:
I will migrate these algorithms from memgraph.py into graph_algorithms dir the same as #130 after that PR is merged

gqlalchemy/memgraph.py
```diff
@@ -543,3 +544,45 @@ def create_relationship(self, relationship: Relationship) -> Optional[Relationsh
...
+
+    These algorithms are integrated into Memgraph codebase and are called
+    within a relationship part of a query. For instance:
+    MATCH p = (:City {name: "Paris"})
+          -[:Road * bfs (r, n | r.length <= 200 AND n.name != "Metz")]->
+          (:City {name: "Berlin"})
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
```
BorisTasevski, line 560: Do we need this constructor?

gqlalchemy/memgraph.py
```diff
@@ -543,3 +544,45 @@ def create_relationship(self, relationship: Relationship) -> Optional[Relationsh
...
+        used e.g. for a filter lambda, to use only relationships of length less
+        than 200:
+            expression="r.length < 200"
+        with the filter lambda being:
+            (r, n | r.length < 200)
+
+        Args:
+            expression: lambda conditions or statements
+        """
+        if expression is None:
```
BorisTasevski, line 585: `return "" if expression is None else f"(r, n | {expression})"`

gqlalchemy/query_builder.py
```diff
@@ -255,13 +262,17 @@ def variable(self) -> str:
     def labels(self) -> str:
         return self._labels if self._labels is not None else ""
 
+    @property
+    def algorithm(self) -> str:
+        return self._algorithm if self._algorithm is not None else ""
```
BorisTasevski, line 267: Can we use positive logic in ifs?

gqlalchemy/query_builder.py
```diff
@@ -579,7 +592,21 @@ def to(
...
+        if algorithm is not None:
+            algorithm_str = str(algorithm)
+        else:
+            algorithm_str = ""
+
+        self._query.append(
+            EdgePartialQuery(
+                variable=variable,
+                labels=type_str,
+                algorithm=algorithm_str,
```
BorisTasevski, line 604: `algorithm = "" if algorithm is None else str(algorithm)`

With this line You can remove if-else above return

gqlalchemy/query_builder.py
```diff
@@ -614,7 +642,21 @@ def from_(
...
+        if algorithm is not None:
+            algorithm_str = str(algorithm)
+        else:
+            algorithm_str = ""
+
+        self._query.append(
+            EdgePartialQuery(
+                variable=variable,
+                labels=labels_str,
+                algorithm=algorithm_str,
```
BorisTasevski, line 654: `algorithm = "" if algorithm is None else str(algorithm)`

With this line You can remove if-else above return

gqlalchemy/memgraph.py
Josipmrden: I think not as well

gqlalchemy/query_builder.py
Josipmrden: I think it's a better practice to use something that we think would pass more often on first if, not on else. That's up for Bruno to decide, and whether people would use or not algorithms. If in doubt, use positive logic.

Josipmrden (changes requested):

gqlalchemy/query_builder.py
brunos252: I changed it to positive logic, I don't think people would use algorithms more often than not, but now it stands out from other @property getters, for variable, labels and properties which are all in negative logic, should we change them also?

Josipmrden (approved):

gqlalchemy/query_builder.py
katarinasupe: If we decide to go with the positive logic, I would change it in all other property getters too.

katarinasupe: Okay, you did that in the newest commit, right? This can then be resolved.

gqlalchemy/memgraph.py
```diff
@@ -543,3 +544,39 @@ def create_relationship(self, relationship: Relationship) -> Optional[Relationsh
...
+    """Abstract class modeling in-Memgraph graph algorithms.
+
+    These algorithms are integrated into Memgraph codebase and are called
+    within a relationship part of a query. For instance:
+    MATCH p = (:City {name: "Paris"})
+          -[:Road * bfs (r, n | r.length <= 200 AND n.name != "Metz")]->
+          (:City {name: "Berlin"})
+    """
+
+    def __str__(self) -> str:
```
BorisTasevski, line 559: Change this to `abstractmethod`

BorisTasevski (changes requested):

g-despot (approved):

BorisTasevski (approved):

PR #167, opened 2022-06-11
[develop < T0106] fix @property being optional when they are not

brunos252:
### Description

After @BorisTasevski noted checking for `None` in a property that is not of type `Optional[...]` is not consistent, I looked at other `@property` uses. 
- for `NodePartialQuery` and `RelationshipPartialQuery` value argument, which is optional, I turned the double negative if-else expression into positive logic, as we previously discussed was easier to understand.
- for arguments `labels` and `properties`, along with `RemovePartialQuery` and `DeletePartialQuery` arguments, I removed the checking and Optional support since they go through their respective `to_cypher_` methods first, which assigns them an empty string if they are `None` to begin with. This is something to discuss, if you think we should still support this for flexibility (e.g. if we were to change `to_cypher_properties` method), let me know.

One other thing I noticed: `RelationshipPartialQuery` receives a `labels` argument, but we agreed on naming the relationship labels as `types`. and also, there must be exactly one so it should be named `type`. Unfortunately this would be a breaking change.

### Pull request type

- [x] Code style update (formatting, renaming)

### Checklist:

- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

- [ ] Tests
- [ ] Code documentation

### DIFF
--- a/gqlalchemy/query_builder.py
+++ b/gqlalchemy/query_builder.py
@@ -245,7 +245,7 @@ class XorNotWhereConditionPartialQuery(WhereNotConditionPartialQuery):
 
 
 class NodePartialQuery(PartialQuery):
-    def __init__(self, variable: str, labels: str, properties: str):
+    def __init__(self, variable: Optional[str], labels: str, properties: str):
         super().__init__(DeclarativeBaseTypes.NODE)
 
         self._variable = variable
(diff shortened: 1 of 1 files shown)


BorisTasevski (changes requested):
Please let's iterate over this one more time on Monday.

BorisTasevski (approved):

katarinasupe (approved):

Issue #195, opened 2022-10-19
[BUG] to_cypher_value badly converts numpy objects


BorisTasevski:
**GQLAlchemy version** 1.3.2

**Describe the bug** When converting NetworkX graph to cypher queries, if the graph is populated with numpy objects the utility method to_cypher_value makes wrong conversions for numpy objects(arrays)

**Expected behavior** numpy array gets converted as array

**Steps to reproduce** 
```
graph = nx.Graph()
graph.add_nodes_from([
    (1, {"Key": np.array(["Value1", "Value2"], dtype=object)}),
]
)

cypher = nx_to_cypher(graph)

print (list(cypher)[0])
```

**Expected result**
`CREATE ( {Key: ['Value1', 'Value2'], id: 1});`

**Actual result**
`CREATE ( {Key: '['Value1' 'Value2']', id: 1});`

katarinasupe:
@BorisTasevski Please add steps to reproduce.

BorisTasevski:
Solved with

https://github.com/memgraph/gqlalchemy/pull/208

PR #208, opened 2022-12-02
added initial support for numpy ndarrays and scalars

BorisTasevski:
### Description

added initial support for numpy ndarrays and scalars

### Pull request type

Please delete options that are not relevant.

- [X] Bugfix

### Related issues

Closes (link to issue)

[to_cypher_value badly converts numpy objects](https://github.com/memgraph/gqlalchemy/issues/195)


katarinasupe:
Does this change need to be added [**here**](https://github.com/memgraph/gqlalchemy/blob/main/gqlalchemy/models.py#L145) too? So that it works when the property of a node is ndarray instead of a list in the class definition?

tests/test_utilities.py
```diff
@@ -37,6 +38,9 @@
         ({"k1": 123, "k2": {"subkey1": "abc"}}, "{k1: 123, k2: {subkey1: 'abc'}}"),
         (None, "null"),
         (True, "True"),
+        (np.array([1, 2, 3]), "[1, 2, 3]"),
+        (np.array(["1", "2", "3"]), "['1', '2', '3']"),
+        (np.array([[1, 2], [3, 4], [4, 5]]), "[[1, 2], [3, 4], [4, 5]]"),
```
Josipmrden, line 43: can you add one for floats?

Josipmrden (commented):
minor comment

BorisTasevski:
> Does this change need to be added [**here**](https://github.com/memgraph/gqlalchemy/blob/main/gqlalchemy/models.py#L145) too? So that it works when the property of a node is ndarray instead of a list in the class definition? Otherwise, all seems fine.

This closes the issue within the transformation module, we can create a separate task to look into the models

katarinasupe:
> > Does this change need to be added [**here**](https://github.com/memgraph/gqlalchemy/blob/main/gqlalchemy/models.py#L145) too? So that it works when the property of a node is ndarray instead of a list in the class definition? Otherwise, all seems fine.
> 
> This closes the issue within the transformation module, we can create a separate task to look into the models

Okay, I opened a new issue for the models, so we don't forget.

Josipmrden (approved):

katarinasupe (approved):

PR #210, opened 2022-12-05
[develop < T0123] Graph project implementation

brunos252:
### Description

Added support for graph project feature while trying to make it easy to use. The idea is to pass a node label and relationship types that you want to use in the subgraph to the `call` method for query modules. GQLAlchemy constructs the query for project and inserts the `graph` argument in the query module call. both can be as many as you like. Given that the labels and types are used to construct the match part of the query, I also included the possibility to use `subgraph_query` argument and insert the Cypher path in string form.

Notes:
- I added the ProjectPartialQuery, which is a MATCH - WITH query, I used `DeclarativeBaseTypes.MATCH` as its type. This and the construction of the MATCH query could have been done in other ways, for instance using QueryBuilder methods, but I did it like this since the `WITH project(p) as graph` is so specific. 
- Also, I changed the previously implemented `PropertyVariable` to `CypherVariable`, this is only here to escape the `to_cypher_value` method used to add quotes around string arguments, so we can use variables. Since the `graph` variable is used in the QM call for the subgraph, this was needed.

### Pull request type

- [x] Feature

### Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

- [ ] Tests
- [ ] Code documentation

### DIFF
--- a/gqlalchemy/query_builders/memgraph_query_builder.py
+++ b/gqlalchemy/query_builders/memgraph_query_builder.py
@@ -31,7 +31,7 @@ from gqlalchemy.query_builders.declarative_base import (  # noqa F401
 )
 from gqlalchemy.vendors.database_client import DatabaseClient
 from gqlalchemy.vendors.memgraph import Memgraph
-from gqlalchemy.utilities import CypherVariable
+from gqlalchemy.utilities import CypherVariable, CypherNode, CypherRelationship, RelationshipDirection
 
 
 class MemgraphQueryBuilderTypes(DeclarativeBaseTypes):
@@ -92,19 +92,10 @@ class QueryBuilder(DeclarativeBase):
         Returns:
             a string representing a MATCH query for a path with given node labels and relationship types.
         """
-        query = " MATCH p="
-        node = "()" if node_label is None else f"(:{node_label.upper()})"
+        node = CypherNode(node_label)
 
-        query += node + "-"
-
-        if isinstance(relationship_types, str):
-            query += f"[:{relationship_types}]"
-        elif isinstance(relationship_types, Iterable):
-            rels = " | :".join(relationship_types)
-            query += f"[:{rels}]"
-
-        # directed is OK only if we have single label, homogenous graphs
-        query += "->" + node
+        # directed is OK only if we have single label, homogenous graphs (current implementation)
+        query = f" MATCH p={node}{CypherRelationship(relationship_types, RelationshipDirection.RIGHT)}{node}"
 
         return query
 
--- a/gqlalchemy/query_builders/memgraph_query_builder.py
+++ b/gqlalchemy/query_builders/memgraph_query_builder.py
@@ -89,8 +89,9 @@ class QueryBuilder(DeclarativeBase):
         """Constructs a MATCH query defining a subgraph using one or two node labels and relationship types.
 
         Args:
-            node_labels: If string, used for both sides of the relationship, if list of two, used
+            node_labels: If string, used for both sides of the relationship, if list of two, used in order.
             relationship_types: A string or iterable of types of relationships used in the subgraph.
+            relationship_direction: Enum representing direction.
 
         Returns:
             a string representing a MATCH query for a path with given node labels and relationship types.
@@ -127,8 +128,9 @@ class QueryBuilder(DeclarativeBase):
               format `query_module.procedure`.
             arguments: A string representing the arguments of the procedure in
               text format.
-            node_label: Label of nodes to be used in the subgraph.
+            node_labels: Labels that define the subgraph.
             relationship_types: Types of relationships to be used in the subgraph.
+            relationship_direction: Direction of the relationship.
             subgraph_query: Optional way to define the subgraph via a Cypher MATCH clause.
 
         Returns:
--- a/gqlalchemy/query_builders/memgraph_query_builder.py
+++ b/gqlalchemy/query_builders/memgraph_query_builder.py
@@ -97,14 +97,14 @@ class QueryBuilder(DeclarativeBase):
             a string representing a MATCH query for a path with given node labels and relationship types.
         """
         if not node_labels or isinstance(node_labels, str):
-            node1 = node2 = CypherNode(node_labels)
+            node1 = second_node = CypherNode(node_labels)
         else:
             if len(node_labels) > 2:
                 raise ValueError("only two node labels are permitted, one for outgoing node and one for incoming node.")
-            node1 = CypherNode(node_labels[0])
-            node2 = CypherNode(node_labels[1])
+            first_node = CypherNode(node_labels[0])
+            second_node = CypherNode(node_labels[1])
 
-        query = f" MATCH p={node1}{CypherRelationship(relationship_types, relationship_direction)}{node2}"
+        query = f" MATCH p={first_node}{CypherRelationship(relationship_types, relationship_direction)}{second_node}"
 
         return query
 
--- a/gqlalchemy/query_builders/memgraph_query_builder.py
+++ b/gqlalchemy/query_builders/memgraph_query_builder.py
@@ -81,21 +81,29 @@ class QueryBuilder(DeclarativeBase):
         return self
 
     def construct_subgraph_query(
-        self, node_label: Optional[str] = None, relationship_types: Optional[Union[str, Iterable[str]]] = None
+        self,
+        node_labels: Optional[Union[str, Iterable[str]]] = None,
+        relationship_types: Optional[Union[str, Iterable[str]]] = None,
+        relationship_direction: Optional[RelationshipDirection] = RelationshipDirection.RIGHT,
     ) -> str:
-        """Constructs a MATCH query defining a subgraph using zero or one node labels and relationship types.
+        """Constructs a MATCH query defining a subgraph using one or two node labels and relationship types.
 
         Args:
-            node_label: A string representing a node label to be used in this subgraph.
+            node_labels: If string, used for both sides of the relationship, if list of two, used
             relationship_types: A string or iterable of types of relationships used in the subgraph.
 
         Returns:
             a string representing a MATCH query for a path with given node labels and relationship types.
         """
-        node = CypherNode(node_label)
+        if not node_labels or isinstance(node_labels, str):
+            node1 = node2 = CypherNode(node_labels)
+        else:
+            if len(node_labels) > 2:
+                raise ValueError("only two node labels are permitted, one for outgoing node and one for incoming node.")
+            node1 = CypherNode(node_labels[0])
+            node2 = CypherNode(node_labels[1])
 
-        # directed is OK only if we have single label, homogenous graphs (current implementation)
-        query = f" MATCH p={node}{CypherRelationship(relationship_types, RelationshipDirection.RIGHT)}{node}"
+        query = f" MATCH p={node1}{CypherRelationship(relationship_types, relationship_direction)}{node2}"
 
         return query
 
@@ -103,8 +111,9 @@ class QueryBuilder(DeclarativeBase):
         self,
         procedure: str,
         arguments: Optional[Union[str, Tuple[Union[str, int, float]]]] = None,
-        node_label: Optional[str] = None,
+        node_labels: Optional[Union[str, Iterable[str]]] = None,
         relationship_types: Optional[Union[str, Iterable[str]]] = None,
+        relationship_direction: Optional[RelationshipDirection] = RelationshipDirection.RIGHT,
         subgraph_query: str = None,
     ) -> "DeclarativeBase":
         """Override of base class method to support Memgraph's subgraph functionality.
@@ -137,8 +146,12 @@ class QueryBuilder(DeclarativeBase):
                     CALL export_util.json(graph, '/home/user')`
         """
 
-        if not (node_label is None and relationship_types is None):
-            subgraph_query = self.construct_subgraph_query(node_label=node_label, relationship_types=relationship_types)
+        if not (node_labels is None and relationship_types is None):
+            subgraph_query = self.construct_subgraph_query(
+                node_labels=node_labels,
+                relationship_types=relationship_types,
+                relationship_direction=relationship_direction,
+            )
 
         if subgraph_query is not None:
             self._query.append(ProjectPartialQuery(subgraph_query=subgraph_query))
--- a/gqlalchemy/query_builders/memgraph_query_builder.py
+++ b/gqlalchemy/query_builders/memgraph_query_builder.py
@@ -97,7 +97,7 @@ class QueryBuilder(DeclarativeBase):
             a string representing a MATCH query for a path with given node labels and relationship types.
         """
         if not node_labels or isinstance(node_labels, str):
-            node1 = second_node = CypherNode(node_labels)
+            first_node = second_node = CypherNode(node_labels)
         else:
             if len(node_labels) > 2:
                 raise ValueError("only two node labels are permitted, one for outgoing node and one for incoming node.")
--- a/gqlalchemy/query_builders/memgraph_query_builder.py
+++ b/gqlalchemy/query_builders/memgraph_query_builder.py
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional, Union, Tuple, Iterable
+from typing import Optional, Union, Tuple, List
+from string import ascii_lowercase
 
 from gqlalchemy.query_builders.declarative_base import (  # noqa F401
     Call,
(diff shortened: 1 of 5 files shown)


gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -79,8 +80,109 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
+        self, node_label: Optional[str] = None, relationship_types: Optional[Union[str, Iterable[str]]] = None
+    ) -> str:
+        """Constructs a MATCH query defining a subgraph using zero or one node labels and relationship types.
+
+        Args:
+            node_label (Optional[str], optional): node label to be used in subgraph
+            relationship_types (Optional[Union[str, Iterable[str]]], optional): types of relationships used in subgraph.
+
+        Returns:
+            str: a MATCH query for a path with given node labels and relationship types
```
katarinasupe, line 93: str is not variable name, but return type, so maybe just:
```suggestion
            A string that represents a MATCH query for a path with given node labels and relationship types.
```

So far, we followed [Google docstring](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings) standard I think, so it is enough to describe the return argument and add a period at the end of each line.

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -79,8 +80,109 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
+        """
+        query = " MATCH p="
+        if node_label is None:
+            node = "()"
+        else:
+            node = f"(:{node_label.upper()})"
```
katarinasupe, line 99:
```suggestion
node = "()" if node_label is None else f"(:{node_label.upper()})"
```

I am writing this directly here in the comments, so not sure about the correctness. Also, this is just a suggestion, there is nothing wrong with the above. 

Another suggestion is to define the node as a Cypher object somewhere. This could be useful later if you're going to add multiple labels. My suggestion is that this `CypherNode` class is always created like `"(" + [":label" for label in labels]?  + ")"`.  Similar could be applied to directed/undirected relationships.

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -79,8 +80,109 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
+        query += node + "-"
+
+        if isinstance(relationship_types, str):
+            query += f"[:{relationship_types}]"
+        elif isinstance(relationship_types, Iterable):
+            rels = " | :".join(relationship_types)
+            query += f"[:{rels}]"
+
+        # directed is OK only if we have single label, homogenous graphs
+        query += "->" + node
```
katarinasupe, line 110: One more thing also poped into my mind - we can create paths in that way. A path is node-relationship-node. If we have these defined, we just append them. That can be used to simplify your code below.

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -79,8 +80,109 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
+
+
+class ProjectPartialQuery(PartialQuery):
+    def __init__(self, subgraph_query):
+        super().__init__(DeclarativeBaseTypes.MATCH)
+        self.query = subgraph_query
+
+    def construct_query(self) -> str:
+        """Constructs a Project partial querty. Given a Match query, adds path identifier if needed
+        And appends the WITH clause."""
```
katarinasupe, line 183:
```suggestion
        and appends the WITH clause."""
```

tests/query_builders/test_memgraph_query_builder.py
```diff
@@ -52,6 +52,36 @@ def test_load_csv_no_header(memgraph):
...
+        ),
+    ],
+)
+def test_call_subgraph_with_query(memgraph, subgraph_query, expected_query):
+    query_builder = QueryBuilder().call("export_util.json", "/home/user", subgraph_query=subgraph_query)
+    with patch.object(Memgraph, "execute", return_value=None) as mock:
+        query_builder.execute()
+    mock.assert_called_with(expected_query)
+
+
```
katarinasupe, line 84: Add tests where you try calling the Memgraph procedure with only `procedure`, `arguments` and `node_label` arguments, and another one with `procedure`, `arguments` and `relationship_types`. This is just to make sure that all of your `if`s above work well, and if you change something in the future that we notice if it fails.

antepusic (commented):
Looks good to me; I’d also second @katarinasupe’s suggestion about paths

brunos252:
Don't mind the failing tests, it is not related to this PR.

Added the classes for Relationships and Nodes. One thing that looks clumsy to me is that i need support for multiple relationship types which are separated by OR ("|"), because of the MATCH clause, but this could be misunderstood when this is used for other purposes, e.g. creating a relationship which must only have one type

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -79,8 +80,97 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
+
+        Returns:
+            a string representing a MATCH query for a path with given node labels and relationship types.
+        """
+        node = CypherNode(node_label)
+
+        # directed is OK only if we have single label, homogenous graphs (current implementation)
+        query = f" MATCH p={node}{CypherRelationship(relationship_types, RelationshipDirection.RIGHT)}{node}"
+
+        return query
```
Josipmrden, line 100: what's the decision then if source node label is different from target node, like in the example here:
https://memgraph.com/docs/mage/how-to-guides/run-a-subgraph-module

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -79,8 +80,97 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
+            Python: `call('export_util.json', '/home/user', "LABEL", ["TYPE1", "TYPE2"]).execute()
+            Cypher: `MATCH p=(:LABEL)-[:TYPE1 | :TYPE2]->(:LABEL) WITH project(p) AS graph
+                    CALL export_util.json(graph, '/home/user')`
+
+            or
+
+            Python: `call('export_util.json', '/home/user', subgraph_query="MATCH (:LABEL)-[:TYPE]->(:LABEL)").execute()
+            Cypher: `MATCH p=(:LABEL)-[:TYPE1]->(:LABEL) WITH project(p) AS graph
+                    CALL export_util.json(graph, '/home/user')`
+        """
```
Josipmrden, line 138: This subgraph_query is not a path

gqlalchemy/query_builders/memgraph_query_builder.py
Josipmrden: Actually the example is invalid though, since I used tweets for everything

gqlalchemy/query_builders/memgraph_query_builder.py
brunos252: this is intentional, `ProjectPartialQuery` checks if there is `p=` (it can be anything instead of p), and if not, adds it, so you can simply write a query you want to define your subgraph, no need to know about our specific implementation, which uses paths.

gqlalchemy/query_builders/memgraph_query_builder.py
brunos252: Added support for two labels. It is even easier to make it support more than two, so I don't see why not. 
But from the usage side, when going for multiple, you would also have to define relationship types for each relationship connecting the nodes, and would need to keep track of the order you input it, so you would pretty much know how to make those in cypher.

gqlalchemy/query_builders/memgraph_query_builder.py
brunos252: Added this, check it out if its OK

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -81,30 +81,39 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
+        node_labels: Optional[Union[str, Iterable[str]]] = None,
+        relationship_types: Optional[Union[str, Iterable[str]]] = None,
+        relationship_direction: Optional[RelationshipDirection] = RelationshipDirection.RIGHT,
     ) -> str:
-        """Constructs a MATCH query defining a subgraph using zero or one node labels and relationship types.
+        """Constructs a MATCH query defining a subgraph using one or two node labels and relationship types.
 
         Args:
-            node_label: A string representing a node label to be used in this subgraph.
+            node_labels: If string, used for both sides of the relationship, if list of two, used
```
katarinasupe: sentence not finished

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -81,30 +81,39 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
             relationship_types: A string or iterable of types of relationships used in the subgraph.
 
         Returns:
             a string representing a MATCH query for a path with given node labels and relationship types.
         """
-        node = CypherNode(node_label)
+        if not node_labels or isinstance(node_labels, str):
+            node1 = node2 = CypherNode(node_labels)
```
katarinasupe, line 100: I would go with `first_node_label`, `second_node_label` , since it's more clear and since we did not use numbers so far. If you have something better in mind, even better.

katarinasupe (changes requested):

gqlalchemy/utilities.py
```diff
@@ -134,9 +140,75 @@ def to_cypher_qm_arguments(arguments: Optional[Union[str, Tuple[Union[str, int,
...
+
+    def __str__(self) -> str:
+        if not self.labels:
+            return "()"
+
+        return "(:" + ":".join(self.labels) + ")"
+
+
+class RelationshipDirection(Enum):
+    """Defines the direction of CypherRelationship object"""
```
katarinasupe, line 171:
```suggestion
    """Defines the direction of CypherRelationship object."""
```

katarinasupe (changes requested):

gqlalchemy/query_builders/memgraph_query_builder.py
katarinasupe: Yep, nice :)

gqlalchemy/query_builders/memgraph_query_builder.py
brunos252: thought not had

gqlalchemy/query_builders/memgraph_query_builder.py
brunos252: it's not just the label, it's the entire node including parentheses, e.g. `(:LABEL)`

katarinasupe: Then `first_node` and `second_node` or leave it as it is 😄 I will approve the PR now.

katarinasupe (approved):

gqlalchemy/query_builders/memgraph_query_builder.py
brunos252: This request has been implemented.

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -79,8 +80,112 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
+
+    def call(
+        self,
+        procedure: str,
+        arguments: Optional[Union[str, Tuple[Union[str, int, float]]]] = None,
+        node_labels: Optional[Union[str, Iterable[str]]] = None,
+        relationship_types: Optional[Union[str, Iterable[str]]] = None,
+        relationship_direction: Optional[RelationshipDirection] = RelationshipDirection.RIGHT,
+        subgraph_query: str = None,
+    ) -> "DeclarativeBase":
```
Josipmrden, line 119: maybe it is better to use always `RelationshipDirection.RIGHT` to simplify it? I'm not sure if we need both since we can only query a subgraph of max 2 entities here

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -79,8 +80,112 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
 
         return self
 
+    def construct_subgraph_query(
+        self,
```
Josipmrden, line 84: should this be private?

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -79,8 +80,112 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
+        Method can be called with node labels and relationship types, both being optional, which are used to construct
+        a subgraph, or if neither is provided, a subgraph query is used, which can be passed as a string representing a
+        Cypher query defining the MATCH clause which selects the nodes and relationships to use.
+
+        Args:
+            procedure: A string representing the name of the procedure in the
+              format `query_module.procedure`.
+            arguments: A string representing the arguments of the procedure in
+              text format.
+            node_labels: Labels that define the subgraph.
```
Josipmrden, line 131: you explain definitely better the `node_labels` parameter in the `construct_subgraph_query` function. perhaps you should reuse it as people will look at this docstring when seeing how to use projections

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -79,8 +80,112 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
+            A `DeclarativeBase` instance for constructing queries.
+
+        Examples:
+            Python: `call('export_util.json', '/home/user', "LABEL", ["TYPE1", "TYPE2"]).execute()
+            Cypher: `MATCH p=(:LABEL)-[:TYPE1 | :TYPE2]->(:LABEL) WITH project(p) AS graph
+                    CALL export_util.json(graph, '/home/user')`
+
+            or
+
+            Python: `call('export_util.json', '/home/user', subgraph_query="MATCH (:LABEL)-[:TYPE]->(:LABEL)").execute()
```
Josipmrden, line 146: is it easier to specify `(:LABEL)-[:TYPE]->(:LABEL)` instead of `MATCH (:LABEL)-[:TYPE]->(:LABEL)`
I guess the intention was not to use cypher so I would reason to ommit this and generate by itself

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -79,8 +80,112 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
+            Cypher: `MATCH p=(:LABEL)-[:TYPE1]->(:LABEL) WITH project(p) AS graph
+                    CALL export_util.json(graph, '/home/user')`
+        """
+
+        if not (node_labels is None and relationship_types is None):
+            subgraph_query = self.construct_subgraph_query(
+                node_labels=node_labels,
+                relationship_types=relationship_types,
+                relationship_direction=relationship_direction,
+            )
```
Josipmrden, line 156: how would I write a subgraph with any relationship between 2 specified entities?

Josipmrden: or vice versa

gqlalchemy/query_builders/memgraph_query_builder.py
```diff
@@ -79,8 +80,112 @@ def load_csv(self, path: str, header: bool, row: str) -> "DeclarativeBase":
...
 class LoadCsv(DeclarativeBase):
     def __init__(self, path: str, header: bool, row: str, connection: Optional[DatabaseClient] = None):
         super().__init__(connection)
         self._query.append(LoadCsvPartialQuery(path, header, row))
+
+
+class ProjectPartialQuery(PartialQuery):
+    def __init__(self, subgraph_query):
+        super().__init__(DeclarativeBaseTypes.MATCH)
+        self.query = subgraph_query
```
Josipmrden, line 182: make it private since it's an objects property?

tests/query_builders/test_memgraph_query_builder.py
```diff
@@ -57,6 +58,67 @@ def test_load_csv_no_header(memgraph):
...
+def test_call_subgraph_with_type(memgraph):
+    relationship_type = "TYPE1"
+    query_builder = QueryBuilder().call("export_util.json", "/home/user", relationship_types=relationship_type)
+    expected_query = " MATCH p=()-[:TYPE1]->() WITH project(p) AS graph CALL export_util.json(graph, '/home/user') "
+    with patch.object(Memgraph, "execute", return_value=None) as mock:
+        query_builder.execute()
+    mock.assert_called_with(expected_query)
```
Josipmrden, line 99: oh I see here is addressed the question from above that I posted. I think we can resolve the issue then :D

Josipmrden (changes requested):
This is good stuff, looking forward to a how to guide as well since I have a feel it could be useful for people to easily query on subgraphs via Python

gqlalchemy/query_builders/memgraph_query_builder.py
brunos252: Yup, both seem a bit bad because users should use Cypher construct `(:LABEL)-[:TYPE]->(:LABEL)` and omit the `MATCH p=` or `p=`, but this seems better. The `subgraph_query` parameter is for the long tail of uses when someone wants, for instance, a path consisting of three nodes to create a subgraph

gqlalchemy/query_builders/memgraph_query_builder.py
brunos252: Since it is a default value, I'd keep it in, for user flexibility. If we hardcode it, it might be more confusing in some sense, because with this users only need to look at the argument list, and have everything at one place.

But I'm not stubborn about this, let me know if I should change it.

Josipmrden: okay if there is a default value then fine, it's okay

gqlalchemy/query_builders/memgraph_query_builder.py
Josipmrden: I don't see any change here or comment

gqlalchemy/query_builders/memgraph_query_builder.py
Josipmrden: Also not resolved, I don't see any change in `node_labels`

Josipmrden (changes requested):
I don't see any changes applied?

antepusic (approved):
I like what I see

Josipmrden (approved):

Josipmrden (approved):

brunos252:
Not really run queries on a subgraph. "Now you can call query modules on a subgraph using graph projections."
Or more plainly: "graph project feature that enables running algorithms on a specific subset of a graph stored in the database." (from fico's article)
 With the link to how to guide we could also link Fico's blog post or some documentation on the graph projections implementation itself, because it is kind of specific so people understand what we're talking about

vpavicic:
If that blog post which describes how we designed and implemented that feature is the best source of information for the feature, we urgently need to add some other information to the docs :D

RN: Now you can execute procedures from query modules on a subgraph using the project feature. Check the steps in the how-to guide. 

We can add a label that links to the blog post in the how-to guide like the one we have in the mage's how to guide :)

brunos252:
Yes, how-to guide is better: https://memgraph.com/docs/mage/how-to-guides/run-a-subgraph-module
I agree with the RN

PR #309, opened 2024-06-24
Add storage mode info

### LINKED ISSUE #239: Add STORAGE MODE query to query builder
Add a new query to the query builder: 
`STORAGE MODE IN_MEMORY_{TRANSACTIONAL|ANALYTICAL};`
`STORAGE MODE ON_DISK_TRANSACTIONAL`

-> https://memgraph.com/docs/memgraph/reference-guide/storage-modes

katarinasupe:
### Description

I added getter and setter for storage modes. I also updated the test setup to set storage mode to in-memory transactional by default. 

### Pull request type

- [x] Feature

### Related issues

Delete section if this PR doesn't resolve any issues. 

Closes #239 

### Checklist:

- [ ] I have made corresponding changes to the documentation @katarinasupe TODO: generate reference guide

- [ ] Tests
- [ ] Code documentation
- [ ] Documentation on [gqlalchemy/docs](https://github.com/memgraph/gqlalchemy/tree/main/docs)

### DIFF
--- a/gqlalchemy/vendors/memgraph.py
+++ b/gqlalchemy/vendors/memgraph.py
@@ -69,7 +69,7 @@ class MemgraphConstants:
     PROPERTIES = "properties"
     UNIQUE = "unique"
 
-    
+
 class MemgraphStorageMode(Enum):
     IN_MEMORY_TRANSACTIONAL = "IN_MEMORY_TRANSACTIONAL"
     IN_MEMORY_ANALYTICAL = "IN_MEMORY_ANALYTICAL"
@@ -78,7 +78,7 @@ class MemgraphStorageMode(Enum):
     def __str__(self):
         return self.value
 
-      
+
 def create_transaction(transaction_data) -> MemgraphTransaction:
     """Create a MemgraphTransaction object from transaction data.
     Args:
--- a/gqlalchemy/vendors/memgraph.py
+++ b/gqlalchemy/vendors/memgraph.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from enum import Enum
 import os
 import sqlite3
 from typing import List, Optional, Union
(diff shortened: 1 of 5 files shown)


gqlalchemy/vendors/memgraph.py
```diff
@@ -432,3 +442,13 @@ def with_power_bi(self) -> "Memgraph":
...
+
+    def get_storage_mode(self) -> str:
+        """Returns the storage mode of the Memgraph instance."""
+        result = self.execute_and_fetch("SHOW STORAGE INFO;")
+        storage_mode_value = next((item["value"] for item in result if item["storage info"] == "storage_mode"), None)
+        return MemgraphStorageMode(storage_mode_value).value
+
+    def set_storage_mode(self, storage_mode: MemgraphStorageMode) -> None:
+        """Sets the storage mode of the Memgraph instance."""
+        self.execute(f"STORAGE MODE {storage_mode};")
```
antepusic, line 454: Given that changing storage modes is not allowed while Memgraph contains data, could you write an integration test for that case and, if necessary, add exception handling here?

antepusic (changes requested):
We should cover the behavior of this feature when Memgraph is in a state that doesn’t allow changing storage modes (see my comment above)

gqlalchemy/vendors/memgraph.py
katarinasupe: Is it true that changing storage mode to on-disk from in-memory is not allowed, while switching from one in-memory storage mode to another is okay?

antepusic: You’re right @katarinasupe (https://memgraph.com/docs/fundamentals/storage-memory-usage#switch-storage-modes)

antepusic (approved):
LGTM

gqlalchemy/vendors/memgraph.py
katarinasupe: I see we have an exception added in case there is data in the db and I try to switch to on disk storage mode:

> Client received query exception: You cannot switch from an in-memory storage mode to the on-disk storage mode when the database contains data. Delete all entries from the database, run FREE MEMORY and then repeat this query. 

I will add this exception handling and a test for that.

antepusic (approved):

Josipmrden (approved):

